PR #9, opened 2024-12-13
Add ADO Pipelines and Sync from ADO repo

MattB-msft:
Resync from main branch as part of rehome to GitHub


rido-min (changes requested):
There is one typo on 

https://github.com/microsoft/Agents-for-net/blob/31e8c238ac06a6f393b9c80486d3de25ea9965fe/src/tests/Microsoft.Agents.Model.Tests/Microsoft.Agents.Model.Tests.csproj#L29

it should be `Core` otherwise the build fails on linux.

Also, once this is fixed, there are 3 test failures.

rido-min (approved):

PR #11, opened 2024-12-13
Add GitHub Action to retrieve Azure DevOps access token and configure…

MattB-msft:
Creating GitHub Invoke of remote PR.


rido-min (changes requested):
I dont understand in which cases we need to get an ADO token. AzPipelines can be triggered by using the Az Pipelines app. And also seems those are failing.

Can you provide more details in the PR description?

MattB-msft:
I am debugging / developing the invoker for Azure Devops.. Iv not tagged anyone for review yet. 
I need to be in PR mode to get the correct behavior for token create.

rido-min (approved):

PR #21, opened 2025-01-02
Add local Linux Build.

MattB-msft:
Adding Local Linux Build to check core build suitability


.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
+name: Linux Local build and test
```
rido-min, line 1: nit: I recommend using simpler names (with no spaces), and potentially matching the yaml file name -> `CI`

rido-min (approved):

PR #35, opened 2025-01-13
POC: unified OAuth flow via ActivityHandler

tracyboehrer:
POC for consolidating OAuth flow on the bot side... ActivityHandler.  This POC subclasses ActivityHandler, and handles the state management and Activities until the user is signed in.

Compare the AuthBot to the original.  This also includes the Teams and SharePoint SSO Middleware functionality so that the bot dev does not have to add that middleware to the Adapter.

Possible next steps if we want to go this direction:
- Likely put this code in ActivityHandler
- Use ConversationState

We could go further...  Make State load/save automatic too.  This avoids the bot writer from having to override OnTurnAsync and saving state.  It would just happen.

### DIFF
--- a/src/libraries/Hosting/AspNetCore/CloudAdapter.cs
+++ b/src/libraries/Hosting/AspNetCore/CloudAdapter.cs
@@ -13,6 +13,7 @@ using Microsoft.Agents.Core.Models;
 using Microsoft.Agents.BotBuilder;
 using Microsoft.Agents.Connector.Types;
 using System.Text;
+using Microsoft.Extensions.Logging.Abstractions;
 
 namespace Microsoft.Agents.Hosting.AspNetCore
 {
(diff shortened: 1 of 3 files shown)


rido-min (commented):
Although I like the idea of subclassing ActivityHandler for oauth, I'm wondering how to reconcile this change when using the TeamsActivityHandler.

This might be a good case to split OAuth flows, as discussed in https://github.com/microsoft/copilot-sdk-for-js/discussions/182

src/libraries/Hosting/AspNetCore/CloudAdapter.cs
```diff
@@ -46,6 +48,7 @@ public CloudAdapter(
         {
             _activityTaskQueue = activityTaskQueue ?? throw new ArgumentNullException(nameof(activityTaskQueue));
             _async = async;
+            _logger = logger ?? NullLogger.Instance;
```
rido-min, line 51: This line fixes a bug I'm investigating, so I think this change belongs to another PR

tracyboehrer: Matt has another fix related to this as well in another PR.  Also, the OnTurnError is broken because for auth issues, the Body on the ErrorResponseException is null.  Which just causes the original exception to be lost because of a null access.

tracyboehrer:
Another consideration is that this implementation is only appropriate for a non-Dialog based bot.  Dialog uses OAuthPrompt to handle signin, which internally uses the OAuthFlow.

PR #51, opened 2025-01-16
Minor CopilotStudioEchoSkill readme & settings updates

tracyboehrer:


src/samples/CopilotStudioEchoSkill/README.md
```diff
@@ -53,20 +54,19 @@ This sample is intended to introduce you to:
           "Settings": {
               "AuthType": "ClientSecret", // this is the AuthType for the connection, valid values can be found in Microsoft.Agents.Authentication.Msal.Model.AuthTypes.  The default is ClientSecret.
```
rido-min, line 55: what is this secret for?

tracyboehrer: For getting a token.  We don't know if we'll need a token for a reply until the Activity arrives.  It is based on Activity.DeliveryMode.  MCS has some logic about when it uses `normal` and when it uses `ExpectReplies`.  All around a better pattern to set it up to handle any DeliveryMode.

MattB-msft (approved):

PR #58, opened 2025-01-17
Updated config to include TenantID in TokenValidation section required by AspNetExtensions.cs

davidyack:
Updated config to include TenantID in TokenValidation section required by AspNetExtensions.cs without this being provided the token validation will fail and the skill can not be added in Microsoft Copilot Studio


tracyboehrer:
@davidyack Thanks!  Since this is still preview, we get some churn here and there.  I noticed that some of the samples are missing this when it would be beneficial to add it due to Single Tenant issuers.

MattB-msft (approved):

PR #62, opened 2025-01-21
Update CI workflow and project files for improved packaging and versioning

MattB-msft:


rido-min (approved):

rido-min:
@MattB-msft to use nbgv in GH Actions you can use https://github.com/marketplace/actions/nerdbank-gitversioning

PR #69, opened 2025-01-27
Added Streaming Messages models and helpers

tracyboehrer:
This does NOT include the functionality to stream OpenAI messages.  That will be a follow up PR with a "StreamingResponse" class to construct the proper Activities in response to OpenAI streamed responses.


MattB-msft (approved):

gurubhg (commented):
Looks good for me.

PR #116, opened 2025-03-01
Add MCP SDK base

madevoge:
WIP initial protocol support.
- Samples include a simple server
- Implements 3 Transports: STDIO, HTTP-over-SSE and HTTP-over-callbacks
- Includes round-trip tests between Client and Server transports to ensure compatibility
- Includes In-Memory defaults with extensibility points to:
   - Providing custom Session Managers
   - Add or Replace Payload handlers
   - Extending with tools
   - Listening/Tapping into Session traffic to enable background jobs
   - Providing custom Factories
   - Adding of Transports
   - PropertyBag for custom Session properties


MattB-msft (approved):
Inital checking Approval.

PR #133, opened 2025-03-14
Streaming Responses and sample

tracyboehrer:
Added streaming messages for handling OpenAI streamed responses (or anything that streams).

- ITurnContext.StreamingResponse
  -  Created at turn start
  -  Type: IStreamingResponse
- AgentApplication will enforce EndStream at turn end if not already done.  Even in the case of an uncaught exception.
- Only supported on WebChat and Teams
  - Non-streaming channels will buffer the stream and send a final message.
- Intermediate messages are sent on a configurable Interval (in milliseconds).
- The final message can be customized by the bot.

```
private async Task OnMessageAsync(ITurnContext turnContext, ITurnState turnState, CancellationToken cancellationToken)
{
    try
    {
        await turnContext.StreamingResponse.QueueInformativeUpdateAsync("Hold on for an awesome poem...", cancellationToken);

        await foreach (StreamingChatCompletionUpdate update in _chatClient.CompleteStreamingAsync(
            "Write a poem about why Microsoft Agents SDK is so great.",
            new ChatOptions { MaxOutputTokens = 1000 },
            cancellationToken: cancellationToken))
        {
            turnContext.StreamingResponse.QueueTextChunk(update?.ToString());
        }
    }
    finally
    {
        await turnContext.StreamingResponse.EndStreamAsync(cancellationToken:cancellationToken);
    }
}
```


src/libraries/BotBuilder/Microsoft.Agents.BotBuilder/App/AgentApplication.cs
```diff
@@ -669,6 +669,11 @@ public async Task OnTurnAsync(ITurnContext turnContext, CancellationToken cancel
             {
                 // Stop the timer if configured
                 StopTypingTimer();
+
+                if (turnContext.StreamingResponse != null && turnContext.StreamingResponse.IsStreamStarted())
+                {
+                    await turnContext.StreamingResponse.EndStreamAsync(30000, cancellationToken).ConfigureAwait(false);
```
MattB-msft: what is the 30K here and should this be config?

MattB-msft (approved):

src/libraries/BotBuilder/Microsoft.Agents.BotBuilder/App/AgentApplication.cs
tracyboehrer: Ya... you're right.  I'll fix that somehow.  I just didn't want to wait forever.

tracyboehrer: I'll add a StreamingResponse.Timeout that EndStream will default to.

PR #143, opened 2025-03-21
Add header propagation functionality

### LINKED ISSUE #176: Allow to pass headers from incoming request to downstream APIs
Teams wants to be able to carry headers over from the incoming request, to all outgoing request.  Specifically the Connector and Token Service requests.

/ @tracyboehrer 

Original issue: https://github.com/microsoft/Agents-for-js/issues/59

sw-joelmut:
Fixes #176 

## Description
This PR adds header propagation functionality, allowing to choose which headers to transition from incoming to outgoing requests.
This implementation uses an Http middleware, a context to share values, and the options to choose which headers to propagate.
Additionally, it will also work with hosted services, as it registers the new request headers into AsyncLocal for the current context by passing the headers via ActivityWithClaims.

### How to use it
Most of the propagation configuration is done internally, so, a way to use this feature is adding the `UseHeaderPropagation` method call to the application builder as follows in the Program.cs file.
```csharp
var app = builder.Build();
app.UseHeaderPropagation();
```

### How it knows which headers to propagate
The header propagation functionality exposes a way to set which headers to propagate by using C# Attributes. Whenever a class uses the `[HeaderPropagation]` attribute, it will be automatically loaded into the header propagation context.
This functionality provides a collection to Add, Override, Append, and Propagate incoming request headers o outgoing ones.

Here is an example:
```csharp
[HeaderPropagation]
internal class HeaderPropagation : IHeaderPropagationAttribute
{
    public static void LoadHeaders(HeaderPropagationEntryCollection collection)
    {
        // Propagate headers to the outgoing request by adding them to the HeaderPropagationEntryCollection.
    }
}
```

## Detailed Changes
- Added HeaderPropagation folder containing all classes of this functionality.
  - Added HeaderPropagationMiddleware which will be used to intercept any incoming request, assigning the headers into a context for later use.
  - Added HeaderPropagationContext which is in charge of maintaining the headers over the current asynchronous control flow using [AsyncLocal](https://learn.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1?view=net-9.0), also, it filters which headers have to save for the propagation by looking at the options configured in the middleware configuration.
- Added UseHeaderPropagation to the application builder extension that lets a customer enable the header propagation functionality.
- Added AddHeaderPropagation as a private application host extension to register the header propagation related classes.
- Added to the HostedActivityService the ability to read the headers from another http context by obtaining the headers from the ActivityWithClaims instance and assigning them to the new header propagation context.

> [!NOTE]
> **Working with Skills:**
> To propagate headers coming from a Root bot, the Skill will also need to register the middleware via UseHeaderPropagation.

## Testing
The following image shows a testing header being propagated.
image


tracyboehrer:
@sw-joelmut Oh nice.  I have a little POC branch, but I like how this uses existing features.  I'll take a look.

sw-joelmut:
Hi @tracyboehrer, we added the C# Attributes to set which headers require propagation to outgoing requests. The implementation is working when we tested it, but after merging with the latest main changes, things changed and may not work for some scenarios (we suspect Skills).

We marked this PR as draft as it is not fully ready to merge with main, and we wanted to know if you like this new approach better, so we can do the final touches and ensure it is working correctly.

Let us know if there are some improvements that we could apply, thanks!

> [!note]
>
> The `app.UseHeaderPropagation()` middleware will still be required in the bot, as it is used to capture all incoming request headers.

sw-joelmut:
Hi @tracyboehrer, we pushed the latest improvements for setting the header to propagate easier, and updated the PR's description to include attributes.

tracyboehrer:
@sw-joelmut Getting back to this.  I'll let you know.

tracyboehrer:
@sw-joelmut Does this also work for the Invoke/ExpectReplies/Streaming handling?  Which doesn't use the queue.

sw-joelmut:
> @sw-joelmut Does this also work for the Invoke/ExpectReplies/Streaming handling? Which doesn't use the queue.

Hi Tracy, [both queue and non-queue handling](https://github.com/microsoft/Agents-for-net/blob/main/src/libraries/Hosting/AspNetCore/CloudAdapter.cs#L179-L195) have the header propagation functionality. We will need to test the Streaming one as it was added around the same time as we created this PR and wasn't taken into account when testing.
We'll add a task to test this.

Thanks!

src/libraries/Core/Microsoft.Agents.Core/HeaderPropagation/HeaderPropagationContext.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    static HeaderPropagationContext()
+    {
+        HeaderPropagationAttribute.LoadHeaders();
+    }
+
+    /// <summary>
+    /// Gets or sets the request headers that will be propagated based on what's inside the <see cref="HeadersToPropagate"/> property.
+    /// </summary>
+    public static IDictionary<string, StringValues> HeadersFromRequest
```
MattB-msft, line 28: @sw-joelmut 
This looks like we are going to have a problem if we are running 2 agents in the same host process or serving 2 endpoints.  or supporting multiple concurrent requests.  
This looks like we will end up causing cross pollination. 

Also how does this get cleaned up?

src/libraries/Core/Microsoft.Agents.Core/HeaderPropagation/HeaderPropagationContext.cs
```diff
@@ -0,0 +1,106 @@
...
+
+namespace Microsoft.Agents.Core.HeaderPropagation;
+
+/// <summary>
+/// Shared context to manage request headers that will be used to propagate them in the <see cref="HeaderPropagationExtensions.AddHeaderPropagation"/>.
+/// </summary>
+public class HeaderPropagationContext()
+{
+    private static readonly AsyncLocal<IDictionary<string, StringValues>> _headersFromRequest = new();
+    private static HeaderPropagationEntryCollection _headersToPropagate = new();
```
tracyboehrer, line 18: I believe this is populated by virtue of the attribute handling, yes?  There is a question about concurrency.  We do something similar with the serializer options init (via attributes).  I added a lock there because I wasn't sure about the exact module loading thread safety.  But I'm not 100% that is required.

src/libraries/Core/Microsoft.Agents.Core/HeaderPropagation/HeaderPropagationAttribute.cs
```diff
@@ -0,0 +1,62 @@
...
+        // And all the ones we currently have loaded
+        foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+        {
+            LoadHeadersAssembly(assembly);
+        }
+    }
+
+    private static void LoadHeadersAssembly(Assembly assembly)
+    {
+        foreach (var type in GetLoadHeadersTypes(assembly))
```
tracyboehrer, line 31: Take a look at `SerializationInitAttribute.GetLoadOnInitTypes`.  We had to add exception handling because some modules failed on this, and the app would crash.  See if this is relevant here too.

src/libraries/Core/Microsoft.Agents.Core/HeaderPropagation/HeaderPropagationContext.cs
ceciliaavila: We added a lock to the Add operations in HeaderPropagationEntryCollection. Thanks!

src/libraries/Core/Microsoft.Agents.Core/HeaderPropagation/HeaderPropagationAttribute.cs
ceciliaavila: Thanks Tracy, we applied the changes.

ceciliaavila:
Hi @MattB-msft, @tracyboehrer,
We added a unit test in `HeaderPropagationMiddlewareTests` to simulate handling multiple HTTP requests and observe how the static _AsyncLocal_ property behaves in this scenario. This test validates that the static property, together with _AsyncLocal_, operates as intended. Since ASP.NET Core encapsulates request information within the HttpContext (as described [here](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/use-http-context?view=aspnetcore-9.0)), and hosted services may run on separate threads, [AsyncLocal](https://learn.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1?view=net-9.0) ensures that data remains isolated within its own context.
Let us know if you have any further questions.
Thanks!

MattB-msft (approved):

MattB-msft:
> Hi @MattB-msft, @tracyboehrer, We added a unit test in `HeaderPropagationMiddlewareTests` to simulate handling multiple HTTP requests and observe how the static _AsyncLocal_ property behaves in this scenario. This test validates that the static property, together with _AsyncLocal_, operates as intended. Since ASP.NET Core encapsulates request information within the HttpContext (as described [here](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/use-http-context?view=aspnetcore-9.0)), and hosted services may run on separate threads, [AsyncLocal](https://learn.microsoft.com/en-us/dotnet/api/system.threading.asynclocal-1?view=net-9.0) ensures that data remains isolated within its own context. Let us know if you have any further questions. Thanks!

Thanks for validating this

PR #158, opened 2025-03-27
AgentToAgent: Added ChannelConversationReference.AgentConversationId

tracyboehrer:
This reference is supplied to OnAgentReply during reply from Agent2 handling.  This is a convenience addition.  The conversationId would be available in the Activity Agent2 sent as well.


src/libraries/Client/Microsoft.Agents.Client/ChannelConversationReference.cs
```diff
@@ -24,5 +24,7 @@ public class ChannelConversationReference
         /// The name of the Agent.
         /// </summary>
         public string AgentName { get; set; }
+
+        public string AgentConversationId { get; set; }
```
MattB-msft, line 28: is there some reason to prepend "Agent" to the ConversationId here?   is there another ConversationID in this object?

tracyboehrer: Because the Agent1 also has a notion of a different conversationId (with ABS or another agent).  Trying to avoid confusion.

tracyboehrer: I really wish we had another word for this since we killed "skill".  "Agent2" is just weird.

tracyboehrer: Let me ask you this:  Is the class name those values are in clear enough?  i.e., For the Agent2 in question, and not something else.

tracyboehrer: Taking another look at this and... ConversationReference and OAuthScope (which really should be "Scope") are in relation to the Agent1 conversation that created the reference.  The "AgentName" and "AgentConversationId" are in relation to Agent2.

If you have a different idea for naming, I'm all ears.

tracyboehrer: Or we can just abandon this PR.  The user can always get it from the received Activity.Conversation.Id.

tracyboehrer: Actually...

`AgentResponseHandler(ITurnContext turnContext, ITurnState turnState, ChannelConversationReference reference, IActivity agentActivity, CancellationToken cancellationToken);`

Is the handler for responses from Agent2.

`ConversationReference` can be pulled from the `turnContext.Activity`.
`AgentConversationId` from `agentActivityConversation.Id`

Which means the only unique values would be the name of the Agent2 that sent it.  So change that signature to just be the Agent2 name instead of that `reference` class.

MattB-msft: Fair enough.. was just curious of we needed to repeat names inside a class ( IE: context should tell you what the subject is )

MattB-msft (approved):

PR #159, opened 2025-03-27
Added AgentApplicationOptions.WelcomeMessage and handling

tracyboehrer:
Adds an "AgentApplication:WelcomeMessage" setting.  If set, a string welcome message is automatically sent on ConversationUpdate Members Added.

Dev can supply a delegate instead (`AddAgentApplicationOptions`) to create something more than a string message.

Or they can define their own OnCovnersationUpdate handler.


src/samples/AgentToAgent/Agent1/appsettings.json
```diff
@@ -1,8 +1,9 @@
 {
-  "AgentApplicationOptions": {
+  "AgentApplication": {
```
MattB-msft, line 2: I see you changing section names here, but not changing it in the loader?

tracyboehrer: It was wrong in the AgentToAgent samples.  That changed long ago.

MattB-msft:
Let's create a Feature to track this and propagate this feature for tracking to other languages.

PR #170, opened 2025-04-01
Added AgentApplicationBuilder

tracyboehrer:


src/libraries/Builder/Microsoft.Agents.Builder/App/AgentApplicationBuilder.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <param name="fileDownloaders"></param>
+        /// <returns></returns>
+        public AgentApplicationBuilder WithFileDownloaders(IList<IInputFileDownloader> fileDownloaders)
+        {
+            Options.FileDownloaders = fileDownloaders;
+            return this;
+        }
+
+        /// <summary>
+        /// Configures authentication for the application.
```
rido-min: this comment does not match the method name. Authorization or Authentication?..

TeamsAI.Application uses WithAuthentication

MattB-msft (approved):

rido-min:
[apologies for replying in a closed PR, could not find a better place]

How is the `AgentApplicationBuilder` intended to be used with DI flows?

Related to this question, seems like the inheritance in the user code eg `public class AuthAgent : AgentApplication` is used to take advantage of DI and primary constructors, but it could easily use composition instead.. no?

/c @tracyboehrer

PR #189, opened 2025-04-08
AgentApplicationOptions and TurnStateFactory

### LINKED ISSUE #187: `turnState` not persisted across turns
When using `OnActivity` the `turnState` is available in the configured handler.

I was expecting the values added in one turn to be available in future turns, however seems those are not persisted when Storage has not been configured

```cs
var agent = new AgentApplication(new AgentApplicationOptions() { StartTypingTimer = true })
    .OnActivity(ActivityTypes.Message, async (turnContext, turnState, cancellationToken) =>
    {
        int counter = turnState.GetValue("counter", () => 0);
        turnState.SetValue("counter", counter++);
        await turnContext.SendActivityAsync(MessageFactory.Text($"Echo [{counter}]: {turnContext.Activity.Text}"), cancellationToken);
    });
```

To avoid this issue we should consider some of these suggestions:

1. If Storage is required (memory storage as a minimum), make this requirement explicit in the AgentApplicationOption ctor (without having to use DI)
2. If storage is not provided, add memory automatically and produce a Warning

tracyboehrer:
Fixes #187

- Require IStorage for AgentApplicationOptions and AgentApplicationBuilder.
  - This creates the default TurnStateFactory
  - Breaking for current users constructing AgentApplicationOptions, but our samples use the IConfiguration version.  So likely limited impact.
- Added additional doc comments for related classes.

```csharp
var storage = new MemoryStorage();

builder.AddAgent(sp =>
{
    var agent = new AgentApplication(new AgentApplicationOptions(storage));

    agent.OnActivity(ActivityTypes.Message, async (turnContext, turnState, cancellationToken) =>
    {
        int count = turnState.Conversation.GetValue<int>("countKey");
        turnState.Conversation.SetValue("countKey", ++count);

        await turnContext.SendActivityAsync($"[{count}] you said: {turnContext.Activity.Text}", cancellationToken: cancellationToken);
    });

    return agent;
});
```


src/libraries/Builder/Microsoft.Agents.Builder/App/AgentApplicationBuilder.cs
```diff
@@ -24,13 +28,35 @@ public class AgentApplicationBuilder
         /// Configures the turn state factory to use for managing the bot's turn state.
```
rido-min: bot?

src/libraries/Builder/Microsoft.Agents.Builder/App/AgentApplicationOptions.cs
```diff
@@ -9,19 +9,73 @@
...
+    /// via <c>IConfiguration</c> values or programmatically.
+    /// </summary>
+    /// <seealso cref="TurnStateFactory"/>
+    /// <seealso cref="UserAuthorizationOptions"/>
     public class AgentApplicationOptions
     {
+        /// <summary>
+        /// Constructs AgentApplicationOptions programmatically.
+        /// <code>
+        ///   var options = new AgentApplicationOptions()
```
rido-min: If IStorage is required for TurnState.. can we require it in this ctor?

src/libraries/Builder/Microsoft.Agents.Builder/App/AgentApplicationBuilder.cs
```diff
@@ -24,13 +28,35 @@ public class AgentApplicationBuilder
...
             Options.TurnStateFactory = turnStateFactory;
             return this;
         }
 
+        /// <summary>
+        /// Configures the turn state factory to use for managing the bot's turn state.
+        /// </summary>
+        /// <param name="storage">The <see cref="IStorage"/> to use with <see cref="TurnState"/>.</param>
+        /// <remarks>
+        /// Not setting the TurnStateFactory would result in non-persisted <see cref="TurnState"/>.  This could
```
rido-min: I would clarify if "non-persisted" means MemoryStorage

src/libraries/Builder/Microsoft.Agents.Builder/App/AgentApplicationOptions.cs
tracyboehrer: Lets chat about this.  We could.  I hadn't thought through it.  My hesitation is that the AgentApplicationBuilder doesn't, and relies on a Fluent call to add it.  I'm open though.

MattB-msft (approved):

PR #193, opened 2025-04-08
Code-first setup of IConnections

### LINKED ISSUE #15: Feature Request: Create more robustness in configuration
Currently when trying to use libraries like MSAL and other parts of the Agents framework it requires us to setup our appsettings.json a certain way with certain settings.  

One example of where this is done is when calling the "AddDefaultMsalAuth" method where you have to send in a configuration section for "MSALConfiguration".

Most of my enterprise level applications have custom configuration code where we inject various configurations from different locations and variations.  Rather then have the libraries pull directly from configuration could we possibly have an option to send in the variables directly?  I find myself having to write code to dynamically create configuration sections on the fly to do translations to just use the libraries.

tracyboehrer:
Fixes #15 

## No appsetting involved

```csharp
// Register IStorage.  For development, MemoryStorage is suitable.
// For production Agents, persisted storage should be used so
// that state survives Agent restarts, and operate correctly
// in a cluster of Agent instances.
builder.Services.AddSingleton<IStorage, MemoryStorage>();

// Add IConnections instance
builder.Services.AddSingleton<IConnections>(sp =>
{
    return new ConfigurationConnections(
        new Dictionary<string, IAccessTokenProvider> 
        {
            {
                "ServiceConnection", new MsalAuth(sp, new ConnectionSettings() 
                { 
                    //ClientId
                    //AuthType
                    //Authority = "https://login.microsoftonline.com/botframework.com",
                    //Scopes = ["https://api.botframework.com/.default" ]
                    //etc
                }
            )}
        },
        [new ConnectionMapItem() { ServiceUrl = "*", Connection = "ServiceConnection"}]
    );
});

// Add factory for ConnectorClient and UserTokenClient creation
builder.Services.AddSingleton<IChannelServiceClientFactory, RestChannelServiceClientFactory>();

// Add the CloudAdapter, this is the default adapter that works with Azure Bot Service and Activity Protocol Agents.
builder.Services.AddCloudAdapter();

// Add the Agent
builder.Services.AddTransient<IAgent>(sp =>
{
    var options = new AgentApplicationOptions()
    {
        TurnStateFactory = () => new TurnState(sp.GetRequiredService<IStorage>())
    };

    return new Echo(options);
});
```


src/libraries/Core/Microsoft.Agents.Authentication/ConnectionSettingsBase.cs
```diff
@@ -12,24 +12,26 @@ namespace Microsoft.Agents.Authentication
...
+                TenantId = msalConfigurationSection.GetValue<string>("TenantId", null);
+                Scopes = msalConfigurationSection.GetSection("Scopes")?.Get<List<string>>();
             }
             else
             {
-                if (configurationSection != null)
+                if (msalConfigurationSection != null)
                 {
-                    throw new ArgumentException($"Authentication configuration section {configurationSection.Key}, not Found.");
+                    throw new ArgumentException($"Authentication configuration section {msalConfigurationSection.Key}, not Found.");
```
MattB-msft: Add Error Wrapper here? ( there is a common function creating exceptions of a type)

src/libraries/Core/Microsoft.Agents.Authentication/ConnectionSettingsBase.cs
```diff
@@ -12,24 +12,26 @@ namespace Microsoft.Agents.Authentication
...
                 }
                 else
                 {
-                    throw new ArgumentNullException(nameof(configurationSection), "No configuration section provided. An authentication configuration section is required to create a connection settings object.");
+                    throw new ArgumentNullException(nameof(msalConfigurationSection), "No configuration section provided. An authentication configuration section is required to create a connection settings object.");
```
MattB-msft: Add Error Wrapper here? ( there is a common function creating exceptions of a type)

src/libraries/Core/Microsoft.Agents.Authentication/ConnectionSettingsBase.cs
tracyboehrer: Yes.  You originally added Errors there, and used a different way, and not throught.  I'll take care of it

src/libraries/Core/Microsoft.Agents.Authentication/ConnectionSettingsBase.cs
tracyboehrer: Samsies

rido-min:
Why IAgent is added as transient instead of singleton?

Will be great to have a test, sample or doc showing how to add all the config sections, `TokenValidation`, `AgentApplicationOptions`, `RootAgent` and even the options used for MSAL.

MattB-msft:
> Why IAgent is added as transient instead of singleton?
> 
> Will be great to have a test, sample or doc showing how to add all the config sections, `TokenValidation`, `AgentApplicationOptions`, `RootAgent` and even the options used for MSAL.

This is an open discussion we are having now.   There are cases where it makes sense for one or the other. 
will continue this as we move forward.

MattB-msft (approved):

PR #207, opened 2025-04-16
User-Agent changes

tracyboehrer:


rido-min:
are you adding the useragent to the MCS client ?

rido-min:
wondering if the RestTeamsConnectorClient needs to be updated with similar changes

MattB-msft (approved):

PR #218, opened 2025-04-22
Add DEBUG block for stack trace output control

MattB-msft:
Introduced a conditional compilation block in `ExceptionParser.cs` for DEBUG mode. 
Added a TODO to implement a setting or environment variable for controlling stack trace output. The stack trace will be appended if it is not null or empty.

### DIFF
--- a/src/libraries/Core/Microsoft.Agents.Core/Errors/ExceptionParser.cs
+++ b/src/libraries/Core/Microsoft.Agents.Core/Errors/ExceptionParser.cs
@@ -68,8 +68,12 @@ namespace Microsoft.Agents.Core.Errors
             sw.AppendLine("TimeStamp: " + DateTime.UtcNow.ToString("o", CultureInfo.InvariantCulture));
             sw.AppendLine("Error: " + message);
             sw.AppendLine($"HelpLink Url: {helpLink}");
+#if DEBUG
+            //TODO:
+            // Update this code to use a setting or environment variable to control the output of the stack trace.
             if (!string.IsNullOrEmpty(stackTrace))
                 sw.AppendLine("Stack Trace: " + stackTrace);
+#endif
             sw.AppendLine("=====================================");
         }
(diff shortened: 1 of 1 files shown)


src/libraries/Core/Microsoft.Agents.Core/Errors/ExceptionParser.cs
```diff
@@ -68,8 +68,12 @@ private static void FormatExceptionMessage(string source, string targetSite, str
             sw.AppendLine("TimeStamp: " + DateTime.UtcNow.ToString("o", CultureInfo.InvariantCulture));
             sw.AppendLine("Error: " + message);
             sw.AppendLine($"HelpLink Url: {helpLink}");
+#if DEBUG
```
rido-min, line 71: I dont think this is the right fix. Many customers will deploy with DEBUG builds, and their stacks will be pushed to the users. 

I will comment those lines out until we have the setting, or reuse IsDevelopment (that is just an envvar)

MattB-msft: As I said in the comment, we will come back an make this configurable. 
IsDevelopment or anything related to Asp.net is not available here, nor is it expected to be here.

rido-min: IsDevelopment is just an envvar: `"ASPNETCORE_ENVIRONMENT": "Development"`

Publishing the Stack Trace to the end user is not good IMHO, and we cannot say: sorry, you did not build on release.

MattB-msft: @rido-min As I said in the previous comment and in the code, 
I will come back and make it a proper setting, this module (where this code is) is the shared module that is used outside of the hosting packages.

PR #228, opened 2025-04-25
Add workload identity as a MSAL option

danielkoek:
Hi, 

I can see that there is no option for a workload identity for agents, we are running on AKS so we would like to add a workload identity instead, this will make it more secure then federation/clientsecrets

### DIFF
--- a/src/libraries/Authentication/Authentication.Msal/Model/ConnectionSettings.cs
+++ b/src/libraries/Authentication/Authentication.Msal/Model/ConnectionSettings.cs
@@ -146,6 +146,16 @@ namespace Microsoft.Agents.Authentication.Msal.Model
                         throw new ArgumentNullException(nameof(Authority), "TenantId or Authority is required");
                     }
                     break;
+                case AuthTypes.WorkloadIdentity:
+                    if (string.IsNullOrEmpty(ClientId))
+                    {
+                        throw new ArgumentNullException(nameof(ClientId), "ClientId is required");
+                    }
+                    if (string.IsNullOrEmpty(Authority) && string.IsNullOrEmpty(TenantId))
+                    {
+                        throw new ArgumentNullException(nameof(Authority), "TenantId or Authority is required");
+                    }
+                    break;
                 default:
                     break;
             }
--- a/src/libraries/Authentication/Authentication.Msal/MsalAuth.cs
+++ b/src/libraries/Authentication/Authentication.Msal/MsalAuth.cs
@@ -225,6 +225,10 @@ namespace Microsoft.Agents.Authentication.Msal
                     }
                     cAppBuilder.WithClientAssertion((AssertionRequestOptions options) => FetchExternalTokenAsync());
                 }
+                else if (_connectionSettings.AuthType == AuthTypes.WorkloadIdentity)
+                {
+                    // No need to do anything
+                }
                 else
                 {
                     throw new System.NotImplementedException();
--- a/src/libraries/Authentication/Authentication.Msal/MsalAuth.cs
+++ b/src/libraries/Authentication/Authentication.Msal/MsalAuth.cs
@@ -227,7 +227,7 @@ namespace Microsoft.Agents.Authentication.Msal
                 }
                 else if (_connectionSettings.AuthType == AuthTypes.WorkloadIdentity)
                 {
-                    // No need to do anything
+                    // No need to do anything more, MSAL will resolve everything for WorkloadIdentity
                 }
                 else
                 {
--- a/src/libraries/Authentication/Authentication.Msal/Model/AuthTypes.cs
+++ b/src/libraries/Authentication/Authentication.Msal/Model/AuthTypes.cs
@@ -10,6 +10,7 @@ namespace Microsoft.Agents.Authentication.Msal.Model
         ClientSecret,
         UserManagedIdentity,
         SystemManagedIdentity,
-        FederatedCredentials
+        FederatedCredentials,
+        WorkloadIdentity
     }
 }


danielkoek:

danielkoek:
> @danielkoek please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

MattB-msft:
Thanks for your suggestion here,  however, we already support this capability via Federated Identity Credentials.  

Does this not work for your situation?

danielkoek:
> @danielkoek Thanks for your suggestion here. However, we already support this capability via Federated Identity Credentials.
> 
> Does this not work for your situation?

Federation and workloads identity are very different, correct me if I am wrong, but I defo don't need a "FederationId", But i also don't need to supply a password OR a cert, you can use an assertion, so i added that now:
As an example:
https://github.com/Azure/azure-workload-identity/blob/main/examples/msal-net/akvdotnet/TokenCredential.cs
Docs about it:
https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview?tabs=dotnet

src/libraries/Authentication/Authentication.Msal/MsalAuth.cs
```diff
@@ -225,6 +228,18 @@ async Task<String> FetchExternalTokenAsync()
...
+                else if (_connectionSettings.AuthType == AuthTypes.WorkloadIdentity)
+                {
+                    cAppBuilder.WithClientAssertion(() =>
+                    {
+                        // read only once every 5 minutes, less heavy for I/O
+                        if (_lastJwtWorkLoadIdentity != null && DateTimeOffset.UtcNow.Subtract(_lastReadWorkloadIdentity) <= TimeSpan.FromMinutes(5)) 
+                            return _lastJwtWorkLoadIdentity;
+                        _lastReadWorkloadIdentity = DateTimeOffset.UtcNow;
+                        _lastJwtWorkLoadIdentity = File.ReadAllText(_connectionSettings.FederatedTokenFile);
+                        return _lastJwtWorkLoadIdentity;
```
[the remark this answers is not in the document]
danielkoek: No need as it is read only most of the time and if there is a race condition it will end up with a few ms of a timestamp diff and the token will stay the same, so it will not "break" the logic

danielkoek:
@MattB-msft
I have checked it on our system now too and we definitely cant use the federated approach, so for now i have used a local assembly that "mocks" this and that works!

tracyboehrer:
@danielkoek We still see this, Daniël.  Considering it, but it may not be until after Build (May 19) that we have space to get to it.

MattB-msft:
@danielkoek 
Yes.. we want to dig more into this before accepting it, As @tracyboehrer said, It will be post build before we get into it..  
We will take a look through this more in depth then. 

pre that, 
can you provide more information about your configuration?  What are you running your container in / hosting so we can set this up and understand the configuration to make sure we account for it in the future.

danielkoek:
> @danielkoek Yes.. we want to dig more into this before accepting it, As @tracyboehrer said, It will be post build before we get into it.. We will take a look through this more in depth then.
> 
> pre that, can you provide more information about your configuration? What are you running your container in / hosting so we can set this up and understand the configuration to make sure we account for it in the future.

Yea no problem, i think i already mentioned it, workload identity: https://azure.github.io/azure-workload-identity/docs/installation/azwi.html

Just normal AKS cluster with a SA setup to have the workload identity loaded in (azwi will do the SA part too), the example used the token file, so that is what i did in my own implementation as well. Our repo for this is actually public: https://github.com/Uniphar/devops-teams-notifications-api, you can see how i "fixed" it there too.

src/libraries/Authentication/Authentication.Msal/Model/ConnectionSettings.cs
```diff
@@ -68,6 +68,11 @@ public ConnectionSettings(IConfigurationSection msalConfigurationSection) : base
         /// </summary>
         public bool SendX5C { get; set; } = false;
 
+        /// <summary>
+        /// Token path used for the workload identity, like the MSAL example for AKS, equal to AZURE_FEDERATED_TOKEN_FILE 
+        /// </summary>
+        public string FederatedTokenFile { get; set; }
```
MattB-msft, line 76: @danielkoek 
We are getting back to this now,

Im not super comfortable with the idea of storing the token as a file on disk. 
are you intending to have a separate process that stores this?

src/libraries/Authentication/Authentication.Msal/Model/ConnectionSettings.cs
```diff
@@ -146,6 +151,20 @@ public void ValidateConfiguration()
...
+                    {
+                        throw new ArgumentNullException(nameof(ClientId), "ClientId is required");
+                    }
+                    if (string.IsNullOrEmpty(Authority) && string.IsNullOrEmpty(TenantId))
+                    {
+                        throw new ArgumentNullException(nameof(Authority), "TenantId or Authority is required");
+                    }
+                    if (string.IsNullOrEmpty(FederatedTokenFile))
+                    {
+                        throw new ArgumentNullException(nameof(FederatedTokenFile), "FederatedTokenFile is required");
```
MattB-msft: if we are going to use this type of tokens store, we also need to do a check for File.Exists.

src/libraries/Authentication/Authentication.Msal/Model/ConnectionSettings.cs
danielkoek: This is part of a workload identity, it does this storing and replacing the token from time to time, all to do with that k8s has access to "replace" this without actually affecting the system itself

src/libraries/Authentication/Authentication.Msal/Model/ConnectionSettings.cs
danielkoek: Yea for sure! did that now!

tracyboehrer:
@danielkoek Has this version been tested in a live Agent?

danielkoek:
> @danielkoek Has this version been tested in a live Agent?
@tracyboehrer We're currently using this specific implementation: https://github.com/Uniphar/devops-teams-notifications-api/blob/main/src/Teams.Notifications.Api/MsalAuthChanged.cs. It's almost a 1:1 copy, with the main differences being how we handle configuration and some minor adjustments to streamline it for our specific use case.

I don't believe there are PR-specific NuGet builds available, so I'm not entirely sure of the best way to test this change. If there's any documentation or guidance on how to test changes from a PR (e.g., using a local build or NuGet feed), I'd be happy to take a look.

MattB-msft:
@danielkoek  we discussed this more internally,  We are going to accept this PR so that you do get credit for it :) and make any changes we need to based on this change. 

thanks again for your contribution and your patience as we worked out some kinks in our processes.

PR #283, opened 2025-05-14
Add Teams Header Propagation sample

sw-joelmut:
This PR depends on https://github.com/microsoft/Agents-for-net/pull/143

## Description
This PR adds a Teams sample capable of capturing incoming headers to later be propagated to outgoing requests based on the HeaderPropagationAttribute configured in the Extensions.Teams project.

## Specific Changes
  - Added HeaderPropagation sample under samples/Teams folder.

## Testing
The following image shows the sample working and the chosen headers being propagated to the MS Teams service.
image


src/samples/Teams/HeaderPropagation/ConversationStateExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Agents.Builder.State;
```
MattB-msft, line 4: Not clear what is going on here, Why do you need this?

MattB-msft (changes requested):

MattB-msft:
@sw-joelmut Im not clear on why we need this sample, or why it is using the teams extensions. 

All that is being show here is to asp.net's built into Header Propagation features. 

Somone requesting this can be pointed to use: https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.headerpropagationapplicationbuilderextensions.useheaderpropagation?view=aspnetcore-9.0

Could you clarify why we should have a specific sample highlighting an asp.net feature?

tracyboehrer:
I suspect we don't need this sample.  However, we need to get with Teams to discover what headers they want carried over, and the User-Agent addition they want.  They used "teamsai-dotnet/x.x.x" in Teams AI, but probably not what we want for the extension.  Which raises the question... Can multiple instances of `[HeaderPropagation]` append User-Agent values?

Unit tests can perform what this sample is doing?

PR #284, opened 2025-05-14
Added HandlingAttachments

tracyboehrer:
Neither Test in WebChat or Emulator appear to work when adding an attachment.  I used Insomnia (or Postman) to send an Activity with an attachment.

```json
{
	"type": "message",
	"channelId": "webchat",
	"activityId": "1",
	"conversation": {
		"id": "1"
	},
	"from": {
		"id": "from",
		"name": ""
	},
	"recipient": {
		"id": "me",
		"name": "me"
	},
	"attachments": [
		{
			"name": "image",
			"contentType": "image/png",
			"contentUrl": "https://devblogs.microsoft.com/microsoft365dev/wp-content/uploads/sites/73/2024/11/word-image-23435-1.png"
		}
	],
	"text": "image attached",
	"deliveryMode": "expectReplies",
	"serviceUrl": "https://webchat.botframework.com/"
}
```


rido-min:
I've been able to see the attachments from WebChat and Teams

src/samples/test-bots/HandlingAttachments/Properties/launchSettings.TEMPLATE.json
rido-min, line 1: probably for another discussion, but having the launchSettings.TEMPLATE seems a good approach for 1) set the port and 2) override the auth settings without adding the appserttings.Development.json (that during deployment will be included in the deployment artifact)

MattB-msft (approved):

tracyboehrer:
> > I've been able to see the attachments from WebChat and Teams
> 
> The file downloaders only come into play when an Activity is _received_ by the agent, not sent. In Teams, this can be tested in two ways: drag-and-drop (which requires a token), and attaching a file.

Correction... there are 3 ways:  drag-and-drop, attach, and paste.

I believe it's paste that requires the token.  Though the way TeamsAttachmentDownloader was written, it just adds to token to all Teams scenarios.

PR #296, opened 2025-05-16
AdaptiveCardInvokeResponseFactory.AdaptiveCard accepts wrong Value content

### LINKED ISSUE #295: AdaptiveCardInvokeResponseFactory.AdaptiveCard accepts wrong type
The `AdaptiveCard` method accepts an `AdaptiveCard` type, which is resolved to `Microsoft.Agents.Builder.App.AdaptiveCards.AdaptiveCard`.  This is the wrong type.  This would have been a porting error where the original type was `AdaptiveCards.AdaptiveCard`.

This should be changed to a string.

tracyboehrer:
Fixes #295

### DIFF
--- a/src/libraries/Builder/Microsoft.Agents.Builder/App/AdaptiveCards/AdaptiveCardInvokeResponseFactory.cs
+++ b/src/libraries/Builder/Microsoft.Agents.Builder/App/AdaptiveCards/AdaptiveCardInvokeResponseFactory.cs
@@ -13,15 +13,15 @@ namespace Microsoft.Agents.Builder.App.AdaptiveCards
         /// <summary>
         /// Returns response with type "application/vnd.microsoft.card.adaptive".
         /// </summary>
-        /// <param name="adaptiveCard">An Adaptive Card.</param>
+        /// <param name="adaptiveCardJson">An AdaptiveCard JSON value.</param>
         /// <returns>The response that includes an Adaptive Card that the client should display.</returns>
-        public static AdaptiveCardInvokeResponse AdaptiveCard(AdaptiveCard adaptiveCard)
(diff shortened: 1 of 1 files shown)


src/libraries/Builder/Microsoft.Agents.Builder/App/AdaptiveCards/AdaptiveCardInvokeResponseFactory.cs
```diff
@@ -13,15 +14,25 @@ public static class AdaptiveCardInvokeResponseFactory
         /// <summary>
         /// Returns response with type "application/vnd.microsoft.card.adaptive".
         /// </summary>
-        /// <param name="adaptiveCard">An Adaptive Card.</param>
+        /// <param name="adaptiveCard">An AdaptiveCard instance.</param>
         /// <returns>The response that includes an Adaptive Card that the client should display.</returns>
-        public static AdaptiveCardInvokeResponse AdaptiveCard(AdaptiveCard adaptiveCard)
+        public static AdaptiveCardInvokeResponse AdaptiveCard(global::AdaptiveCards.AdaptiveCard adaptiveCard)
```
MattB-msft: the global:: is usually a bad sign.. what's going on here?

tracyboehrer: We have both a namespace and type collision.  What we want there is AdaptiveCards.AdaptiveCard (from the AdaptiveCard package).  But what we are in is Microsoft.Agents.Builder.App.**AdapativeCards** and there is an **AdaptiveCard** type.

Is there another way?  Drop that overload?

MattB-msft (approved):

PR #303, opened 2025-05-21
Fix: Ensure EntityConverter delegates to AIEntityConverter

### LINKED ISSUE #298: AIEntityConverter Not Invoked When Serializing AIEntity Inside Activity.Entities
## Version
- Nuget package version: [1.1.91-beta]]

**Describe the bug**
When serializing an Activity that contains an AIEntity in its Entities list, the EntityConverter is used for AIEntity instead of delegating to the registered AIEntityConverter. As a result, special fields like @type, @context, and @id are not written correctly. 
**To Reproduce**
Test:
```
        [Fact]
        public void ActivityWithDerivedEntitySerializationTest()
        {
            var jsonIn = "{\"membersAdded\":[],\"membersRemoved\":[],\"reactionsAdded\":[],\"reactionsRemoved\":[],\"attachments\":[],\"entities\":[{\"@type\":\"Message\",\"@context\":\"https://schema.org\",\"@id\":\"\",\"additionalType\":[\"AIGeneratedContent\"],\"citation\":[],\"type\":\"https://schema.org/Message\"}],\"listenFor\":[],\"textHighlights\":[]}";

            var activity = ProtocolJsonSerializer.ToObject<Activity>(jsonIn);
            var jsonOut = ProtocolJsonSerializer.ToJson(activity);

            Assert.Equal(jsonIn, jsonOut);
        }
```
Outcome:
```
    Assert.Equal() Failure: Strings differ
                                  ↓ (pos 112)
Expected: ···"s":[],"entities":[{"@type":"Message","@co"···
Actual:   ···"s":[],"entities":[{"atType":"Message","at"···
                                  ↑ (pos 112)
```

msftshlomiyosef:
Fixes #298 
Ensure EntityConverter checks if the value is an AIEntity and delegates serialization, so @type, @context, and @id are included.

### DIFF
--- a/src/libraries/Core/Microsoft.Agents.Core/Serialization/Converters/EntityConverter.cs
+++ b/src/libraries/Core/Microsoft.Agents.Core/Serialization/Converters/EntityConverter.cs
@@ -45,6 +45,16 @@ namespace Microsoft.Agents.Core.Serialization.Converters
             return entity;
         }
 
+        public override void Write(Utf8JsonWriter writer, Entity value, JsonSerializerOptions options)
+        {
+            if (value is AIEntity aiEntity)
+            {
+                JsonSerializer.Serialize(writer, aiEntity, options); // delegates to AIEntityConverter
+                return;
+            }
+            base.Write(writer, value, options);
+        }
+
         protected override void ReadExtensionData(ref Utf8JsonReader reader, Entity value, string propertyName, JsonSerializerOptions options)
         {
             var extensionData = JsonSerializer.Deserialize<JsonElement>(ref reader, options);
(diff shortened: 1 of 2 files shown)


tracyboehrer:
Does https://github.com/microsoft/Agents-for-net/pull/305 replace this PR?

msftshlomiyosef:
> Does #305 replace this PR?

This project solves a very specific problem of serializing AIEntity when the compile-time type is Entity, while preserving the approach of manually handling custom property names in the converters.
#305 solves a broader problem but changes this approach, so it can replace this PR if we agree to adopt the new model.

msftshlomiyosef:
Closing this, as it will be handled here: https://github.com/microsoft/Agents-for-net/pull/305

PR #328, opened 2025-06-10
Handling non-JWT token during OAuth by not caching

### LINKED ISSUE #325: Cannot use non JWT tokens from ABS Token Service
There are OAuth providers that do not use JWT, however this SDK assumes all tokens are JWT

This assumption makes impossible to use another provider such as GitHub

```
Microsoft.IdentityModel.Tokens.SecurityTokenMalformedException
  HResult=0x80070057
  Message=IDX12741: JWT must have three segments (JWS) or five segments (JWE).
  Source=System.IdentityModel.Tokens.Jwt
  StackTrace:
   at System.IdentityModel.Tokens.Jwt.JwtSecurityToken..ctor(String jwtEncodedString)
   at Microsoft.Agents.Connector.RestClients.UserTokenRestClient.AddTokenResponseToCache(String cacheKey, TokenResponse tokenResponse) in C:\_code\Agents-for-net\src\libraries\Client\Microsoft.Agents.Connector\RestClients\UserTokenRestClient.cs:line 425

```

https://github.com/microsoft/Agents-for-net/blob/c07797f615eafe1a358780dab143ab9b32d21094/src/libraries/Client/Microsoft.Agents.Connector/RestClients/UserTokenRestClient.cs?plain=1#L425

tracyboehrer:
Fixes #325


src/libraries/Client/Microsoft.Agents.Connector/RestClients/UserTokenRestClient.cs
```diff
@@ -422,14 +422,35 @@ private static void AddTokenResponseToCache(string cacheKey, TokenResponse token
...
-                tokenResponse.IsExchangeable = IsExchangeableToken(jwtToken);
-
                 if (tokenResponse.Expiration == null)
                 {
-                    // Token Service isn't returning Expiration in TokenResponse
-                    tokenResponse.Expiration = jwtToken.ValidTo;
+                    // Token Service isn't returning Expiration in TokenResponse.  Try to
+                    // get from token.
+
+                    if (tokenResponse.Token.StartsWith("gho_"))
```
rido-min: this seems problematic, I'd say the user token client should be agnostic to the token format. Otherwise, we will need to check all providers to ensure we understand the token format.

Instead, I would add a "TryDecodeJsonWebToken" (if it's not JWT IsExchangable will be false)

tracyboehrer: Not a huge fan of that either.  What we are also doing it trying to get expiration.

tracyboehrer: The only other option is to not cache for non-JWT.

rido-min: I would cache token as strings. The expiration will require more thought imho.

tracyboehrer: The token response is cached.  Which has the string token.

tracyboehrer: See if last commit helps:

If not JWT, still cache.  Callers (like UserAuthoriztion.GetTurnToken) won't be able to do expiration calcs, but it will be available in the user token client cache for the during of the sliding expiration.

We could choose not to cache at all in this case, which would just mean it would make a call to Token Service each time.

tracyboehrer: On further reflection, opted to not cache tokens without expiration info.   Since I've never seen this come across in the TokenResponse from the Token Service, this means non-JWT tokens won't be cached by the UserTokenClient (but will continue to be by UserAuthorization for the duration of the turn).

The practical effect is the non-JWT tokens (without a supplied TokenResponse.Expiration) will just require a Token Service call each time since it's a complete guess otherwise.

I'll confer with ABS about why they don't populate expiration info in the response.  Though in the case of GH tokens, it may just not be available.

PR #335, opened 2025-06-23
Fix Agent-To-Agent communication by retrieving claims from JWT token in ChannelApiController

ceciliaavila:
## Description
This PR fixes an issue in Agent-To-Agent communication. Agent1 found an HTTP 401 error while attempting to forward Agent2's message to the channel, as the Authorization header was not populated correctly.

## Detailed Changes
- Added a `GetClaimsIdentity` method in `HttpHelper` class to retrieve the _claimsIdentity_ from the JWT token when Auth is not configured.
- Call the new method in `ChannelApiController` methods.
- Replace the code in `CloudAdapter's ProcessAsync` to call the new method instead.

## Testing
These images show the Agent-To-Agent sample working with anonymous authentication as well as using credentials.
image


tracyboehrer:
@ceciliaavila This is for when AspNet auth has not been enabled, correct?  In which case, incoming responses should follow the same logic as on /api/messages.  

Test the anonymous route.  That is, hit Agent1 with Emulator (with no appId or secret).

Funny thing is I did the same thing in another branch just with a different name in HttpHelper.  It's fine, I can resolve the other branch later.

tracyboehrer:
@ceciliaavila Or, I should say... test from "Test In WebChat" (which will arrive with a JWT token, and test from Emulator (which is anonymous because no token).

ceciliaavila:
> ame logic as on /api/messages.
> 
> Test the anonymous route. That is, hit Agent1 with Emulator (with no appId or secret).

Hi Tracy,
Anonymous was working. The fix is for communication using credentials. We tested both scenarios after applying this fix.

tracyboehrer (approved):

PR #367, opened 2025-07-11
Added IAccessTokenProvider.GetTokenCredential

### LINKED ISSUE #355: Use IConnection for acquiring Azure TokenCredential
The driving scenario is for using TokenCredential for BlobsStorage or CosmosDbStorage.  But this could be used for any service requiring an Azure TokenCredential.

The POC is:

```csharp
builder.Services.AddSingleton<IStorage>(sp =>
{
    return new BlobsStorage(
        new Uri("https://trboehreasdktstorage.blob.core.windows.net/agent-state"),
        new ConnectionsTokenCredential(sp.GetService<IConnections>().GetConnection("ServiceConnection")));
});
```

With a wrapper TokenCredential class of

```csharp
class ConnectionsTokenCredential(IAccessTokenProvider connection) : TokenCredential
{
    public override AccessToken GetToken(TokenRequestContext requestContext, CancellationToken cancellationToken)
    {
#pragma warning disable CA2012
        return GetTokenAsync(requestContext, cancellationToken).Result;
#pragma warning restore CA2012
    }

    public override async ValueTask<AccessToken> GetTokenAsync(TokenRequestContext requestContext, CancellationToken cancellationToken)
    {
        var token = await connection.GetAccessTokenAsync(null, requestContext.Scopes);
        return new AccessToken(token, DateTimeOffset.UtcNow + TimeSpan.FromMinutes(15));
    }
}```

### PR DESCRIPTION
**Pull request by tracyboehrer**
Fixes #355

Example:
```csharp
builder.Services.AddSingleton<IStorage>(sp =>
{
    // Get a TokenCredential from your defined Connections
    var tokenCredential = sp.GetService<IConnections>().GetConnection("ServiceConnection").GetTokenCredential();

    return new BlobsStorage(
        new Uri("{{your-blobs-storage-endpoint}}/agent-state"),
        tokenCredential);
});
```

### DIFF
--- a/src/libraries/Authentication/Authentication.Msal/MsalAuth.cs
+++ b/src/libraries/Authentication/Authentication.Msal/MsalAuth.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using Azure.Core;
 using Microsoft.Agents.Authentication.Msal.Model;
 using Microsoft.Agents.Authentication.Msal.Utils;
 using Microsoft.Agents.Core;
@@ -16,7 +17,6 @@ using Microsoft.IdentityModel.LoggingExtensions;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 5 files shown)


src/libraries/Authentication/Authentication.Msal/MsalAuth.cs
```diff
@@ -96,7 +102,7 @@ public async Task<string> GetAccessTokenAsync(string resourceUrl, IList<string>
 
                     if (!string.IsNullOrEmpty(accessToken))
```
MattB-msft, line 103: if we are shifting to this authResultFromCache.MsalAuthResul we should do the null check here.

src/libraries/Authentication/Authentication.Msal/MsalTokenCredential.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Azure.Core;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Agents.Authentication.Msal
```
MattB-msft, line 8: I had thought we were going to have this in Builder as we already have Azure.Core there?

PR #447, opened 2025-09-16
Asssembly-level SerializationInitAttribute

tracyboehrer:


src/libraries/Core/Microsoft.Agents.Core/Serialization/SerializationInitAttribute.cs
```diff
@@ -2,83 +2,12 @@
...
-using System.Diagnostics;
-using System.Linq;
-using System.Reflection;
 
 namespace Microsoft.Agents.Core.Serialization
 {
-    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, Inherited = false)]
-    public class SerializationInitAttribute : Attribute
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
+    public class SerializationInitAttribute(Type type) : Attribute
```
olstakh, line 9: May be this attribute can remain, and instead new assembly-level attribute be added? The init logic then can use assembly-level lookup if it's present on the assembly - otherwise it can default back to this one. It'll also be more flexible for source generation if needed, without it being a breaking change for consumers

olstakh: https://github.com/microsoft/Agents-for-net/pull/449

PR #455, opened 2025-09-22
Lower code analysis package version

olstakh:
Using the lowest version that allows current source generator functionality to operate, to cover the most consumers

Version mappings: https://learn.microsoft.com/en-us/visualstudio/extensibility/roslyn-version-support?view=vs-2022


tracyboehrer (approved):

tracyboehrer:
It would appear that the package version was lowered elsewhere.  No effective changes in this PR now.  Closing.

PR #489, opened 2025-10-23
Relax Recipient.Id requirement for MCS

tracyboehrer:


cleemullins (dismissed):
I would suggest leaving in the logging. Handy stuff, that. 

LowWarning "Channel missing the recipient and id field". :)

PR #593, opened 2026-01-06
Add Entity assembly attribute for serialization support

### LINKED ISSUE #591: Regressed type loading delay in Entity serialization init
- Caused by a call to [GetTypes](https://github.com/microsoft/Agents-for-net/blob/main/src/libraries/Core/Microsoft.Agents.Core/Serialization/SerializationInitAssemblyAttribute.cs#L41)

tracyboehrer:
Fixes #591 

In order to avoid using `Assembly.GetTypes` which can cause delays, implemented a similar technique as `SerializationInitAssemblyAttribute` to add assembly attributes for custom Entities.

- Added `EntityInitSourceGenerator` which will looks for subclasses of `Entity`.  If found adds an assembly attribute `EntityInitAssemblyAttribute`.
- This has the same behavior as before
  - If the Entity subclass includes `EntityNameAttribute`, that name will be used
  - Otherwise, the class name will be used

### DIFF
--- a/src/libraries/Core/Microsoft.Agents.Core.Analyzers/Helpers/HashCode.cs
+++ b/src/libraries/Core/Microsoft.Agents.Core.Analyzers/Helpers/HashCode.cs
@@ -1,6 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
+﻿
 using System.ComponentModel;
 using System.Runtime.CompilerServices;
 using System.Security.Cryptography;
--- a/src/libraries/Core/Microsoft.Agents.Core.Analyzers/Helpers/ObjectImmutableArraySequenceEqualityComparer.cs
+++ b/src/libraries/Core/Microsoft.Agents.Core.Analyzers/Helpers/ObjectImmutableArraySequenceEqualityComparer.cs
@@ -1,6 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
+﻿
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;


src/libraries/Core/Microsoft.Agents.Core.Analyzers/EntityInitSourceGenerator.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public void Initialize(IncrementalGeneratorInitializationContext context)
+        {
+            // Step 1: Get all class declarations in the compilation
+            var classDeclarations = context.SyntaxProvider
+                .CreateSyntaxProvider(
+                    predicate: static (node, _) => node is ClassDeclarationSyntax,
+                    transform: static (ctx, _) => (ClassDeclarationSyntax)ctx.Node
+                )
+                .Where(static cls => cls != null);
```
olstakh: I suggest not collecting `ClassDeclarationSyntax`, because it's not cacheable, and the generator will run much more than it needs to. Consider collecting what is precisely needed from the class declarations, in order to produce the code

src/libraries/Core/Microsoft.Agents.Core/Serialization/ProtocolJsonSerializer.cs
```diff
@@ -32,17 +33,28 @@ public static class ProtocolJsonSerializer
...
         {
+            EntityTypes = new();
+            EntityTypes[Models.EntityTypes.ActivityTreatment] = typeof(ActivityTreatment);
+            EntityTypes[Models.EntityTypes.AICitation] = typeof(AIEntity);
+            EntityTypes[Models.EntityTypes.GeoCoordinates] = typeof(GeoCoordinates);
+            EntityTypes[Models.EntityTypes.Mention] = typeof(Mention);
+            EntityTypes[Models.EntityTypes.Place] = typeof(Place);
+            EntityTypes[Models.EntityTypes.ProductInfo] = typeof(ProductInfo);
+            EntityTypes[Models.EntityTypes.StreamInfo] = typeof(StreamInfo);
+            EntityTypes[Models.EntityTypes.Thing] = typeof(Thing);
```
olstakh: Consider not initializing public property implicitly while initializing another public property explicitly.

This is for `SerializationOptions`, i'd suggest having another method for `EntityTypes` that returns a dictionary and is assigned, instead of implicitly mutating the property

src/libraries/Core/Microsoft.Agents.Core.Analyzers/EntityInitSourceGenerator.cs
tracyboehrer: Check it out now

src/libraries/Core/Microsoft.Agents.Core.Analyzers/EntityInitSourceGenerator.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            // Step 2: resolve Entity type
+            var myTypeProvider = compilationProvider
+                .Select(static (compilation, _) => compilation.GetTypeByMetadataName(EntityTypeFullName));
+
+            // Step 3: find derived types
+            var derivedTypesProvider =
+                compilationProvider
+                    .Combine(myTypeProvider)
+                    .Select(static (pair, ct) => FindDerivedTypes(pair.Left, pair.Right));
```
olstakh: I'd suggest returning array of strings, instead of `INamedSymbol`. Also add `.WithComparer` and provide the value-type array comparer

i.e.
```csharp
.WithComparer(new ObjectImmutableArraySequenceEqualityComparer());
```

Plus, i'd assumed you have to `.Collect` the data into a single batch, although may be im mistaken

tracyboehrer: What is the advantage?

olstakh: Caching, so Roslyn doesn't run generator unnecessarily. Two different arrays of strings are different by default, even if elements are the same. Because array is a reference type. With that custom comparer - arrays will be considered the same if they contain the same elements, so generator won't be re-run in such case

As for `INamedSymbol` - same principle. At the moment it's a reference type, but if we get strings from it - they will be compared as values

MattB-msft (approved):

PR #714, opened 2026-02-12
Add new WriteAsync condition and ETag response result

sw-joelmut:
## Description
This PR extends the functionality of the WriteAsync methods in Blobs, Cosmos, and Memory storages by allowing users to pass new options that currently contain the ifNotExists condition, and to return the created/updated ETag as a result of the operation.
The purpose of these changes is to introduce these improvements without modifying the existing signature of the IStorage interface.

There are multiple approaches to accomplish this, but the least hacky way was to add the new WriteAsync overload method to each Storage without impacting the interface or creating a second IStorageExt with new signatures.
Here are two of the most relevant alternatives (POC) to the current changes in the PR, without adding more WriteAsync overloads:
image

### AsyncLocal + Disposable options
1. Instantiate a write options context containing the configuration options the WriteAsync will use, while also having the logic to save the options in the AsyncLocal context (non-static) and dispose them when necessary.
2. Perform the storage.WriteAsync as we normally do, and internally it gets the current context options.
3. Internally when the storage write operation returns the ETag, it assigns it to the write options context.

### StoreItem wrapper
1. Instantiate the store item wrapper by setting the options and the data to write in the storage.
2. Share the wrapper to the storage.WriteAsync method as store information.
3. The WriteAsync will detect the type is a StoreItem wrapper, extract the options and data separately.
4. Internally when the storage write operation returns the ETag, it assigns it to the StoreItem wrapper for later consumption.

### Other less fisable alternatives
image

## Testing
Once the PR is no longer in draft, we will include tests of Agent samples working with the storages.

### DIFF
--- a/src/libraries/Storage/Microsoft.Agents.Storage.Blobs/BlobsStorage.cs
+++ b/src/libraries/Storage/Microsoft.Agents.Storage.Blobs/BlobsStorage.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Text.Json;
 using System.Text.Json.Nodes;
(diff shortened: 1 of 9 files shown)


tracyboehrer:
I think short of a v2, we should just have an extended interface for IStorage.  As Matt suggested extend the existing with the new operations.  There is possibly another operation about getting a paginated list of records, assuming we can abstract that across providers.

sw-joelmut:
Great!, thanks for the details. We’ll plan the implementation using the information you provided. One suggestion i want to mention is that maybe we should consider introducing a separate IStorageV2 so we don’t have different method signatures when extending IStorage to maintain.
image

ceciliaavila:
Replaced by #883

PR #728, opened 2026-02-25
Add conditions to choose ConnectorClient scope

sw-joelmut:
Fixes #723

## Description
This PR adds a condition to only use the `azp` and `appid` JWT identity values when the `Activity.Recipient.Role` is `Skill`, otherwise will resolve `api.botframework.com` as scope.

## Testing
The following image shows the Agent to Agent operations using MSTeams and WebChat
image


src/libraries/Builder/Microsoft.Agents.Builder/ChannelServiceAdapterBase.cs
```diff
@@ -282,6 +283,27 @@ public override async Task<InvokeResponse> ProcessActivityAsync(ClaimsIdentity c
...
 
+        /// <summary>
+        /// Resolves the appropriate authentication scope for the connector client based on the identity and activity.
+        /// </summary>
+        /// <param name="claimsIdentity">A <see cref="ClaimsIdentity"/> for the conversation.</param>
+        /// <param name="activity">The activity to process.</param>
+        /// <returns>The resolved authentication scopes for the connector client.</returns>
+        private static IList<string> ResolveConnectorClientScopes(ClaimsIdentity claimsIdentity, IActivity activity)
+        {
+            IList<string> defaultScopes = [$"{AuthenticationConstants.BotFrameworkScope}/.default"];
```
MattB-msft, line 294: What is Blueprint based?   ( may not be a BotFramework API Scope )

sw-joelmut: I see that when we execute `ChannelServiceFactory.CreateConnectorClientAsync` we pass the scopes that `ResolveConnectorClientScopes` resolves, but in [`CreateConnectorClientAsync`](https://github.com/microsoft/Agents-for-net/blob/main/src/libraries/Builder/Microsoft.Agents.Builder/RestChannelServiceClientFactory.cs#L111) we only use scopes for non-Agentic requests.
Is this what you are refering to, or am i missing something?

Thanks for mentioning!

tracyboehrer:
Closing this for now.  Keep the branch though.

PR #811, opened 2026-05-05
Refactor AzureBotUserAuthorization to not rely on exceptions as a signal

sw-joelmut:
Fixes #717

This pull request refactors the OAuth flow handling in the authentication system to use a new structured result type (`OAuthFlowResult`) instead of exceptions and nulls for flow control. This makes the code more robust, easier to understand, and improves testability. The main changes include introducing the `OAuthFlowResult` and `OAuthFlowStatus` types, updating the flow logic to use these, and updating tests to verify the new behavior.

**OAuth flow result handling improvements:**

* Introduced the `OAuthFlowResult` and `OAuthFlowStatus` types to represent the outcome of an OAuth flow in a structured way, replacing the use of exceptions and nulls for flow control. (`src/libraries/Builder/Microsoft.Agents.Builder/UserAuth/TokenService/OAuthFlowResult.cs`)
* Updated `OAuthFlow.ContinueFlowAsync` and related methods to return an `OAuthFlowResult` instead of throwing exceptions or returning null, and refactored the logic in `RecognizeTokenAsync` to return appropriate status values. (`src/libraries/Builder/Microsoft.Agents.Builder/UserAuth/TokenService/OAuthFlow.cs`) [[1]](diffhunk://#diff-7fbb0a3e3e0beddeb752391fc8b4b8d101387cd2bfd33ca4622a91872239ac08L72-R77) [[2]](diffhunk://#diff-7fbb0a3e3e0beddeb752391fc8b4b8d101387cd2bfd33ca4622a91872239ac08L101-R100) [[3]](diffhunk://#diff-7fbb0a3e3e0beddeb752391fc8b4b8d101387cd2bfd33ca4622a91872239ac08L243-R245) [[4]](diffhunk://#diff-7fbb0a3e3e0beddeb752391fc8b4b8d101387cd2bfd33ca4622a91872239ac08L268-R268) [[5]](diffhunk://#diff-7fbb0a3e3e0beddeb752391fc8b4b8d101387cd2bfd33ca4622a91872239ac08L406-R403)

**Dialog and authorization logic updates:**

* Refactored `OAuthPrompt.ContinueDialogAsync` and `AzureBotUserAuthorization.OnContinueFlow` to use the new `OAuthFlowResult` structure, handling timeouts and user cancellations explicitly via status codes instead of exceptions. (`src/libraries/Builder/Microsoft.Agents.Builder.Dialogs/Prompts/OAuthPrompt.cs`, `src/libraries/Builder/Microsoft.Agents.Builder/UserAuth/TokenService/AzureBotUserAuthorization.cs`) [[1]](diffhunk://#diff-1f5368a8c5531912d099ec5f37f1506dbc1294ce3c72f1c040dcd1a1f8edf76fL182-R196) [[2]](diffhunk://#diff-1f5368a8c5531912d099ec5f37f1506dbc1294ce3c72f1c040dcd1a1f8edf76fL238-L243) [[3]](diffhunk://#diff-e10e7b360009c409937f224c65f67a377b0d41aa2354039db585b5be3980e39cL206-R220) [[4]](diffhunk://#diff-e10e7b360009c409937f224c65f67a377b0d41aa2354039db585b5be3980e39cL238-L240)
* Removed the now-unnecessary `UserCancelledException` class from the codebase, as user cancellation is now represented by a status code. (`src/libraries/Builder/Microsoft.Agents.Builder/UserAuth/TokenService/UserCancelledException.cs`, `src/libraries/Builder/Microsoft.Agents.Builder.Dialogs/Prompts/OAuthPrompt.cs`) [[1]](diffhunk://#diff-aecf4611e7b81aedc5d714f02623328be04299bb33e19effdd99d3065ba793caL1-L11) [[2]](diffhunk://#diff-1f5368a8c5531912d099ec5f37f1506dbc1294ce3c72f1c040dcd1a1f8edf76fR317-R320)

**Test updates:**

* Updated and expanded unit tests in `OAuthFlowTests` to verify the new result-based flow, including cases for pending, timed out, user cancelled, and successful token retrieval scenarios. (`src/tests/Microsoft.Agents.Builder.Tests/OAuthFlowTests.cs`) [[1]](diffhunk://#diff-aa962826b3e49fe77bd02898747d1b55124747eb300ae6bcef24cffe44175319L100-R100) [[2]](diffhunk://#diff-aa962826b3e49fe77bd02898747d1b55124747eb300ae6bcef24cffe44175319L130-R130) [[3]](diffhunk://#diff-aa962826b3e49fe77bd02898747d1b55124747eb300ae6bcef24cffe44175319L168-R260)

These changes modernize the OAuth flow logic, making it more maintainable and predictable.

The following images show the AzureBotUserAuthorization and oAuthPrompt working.
image
image


src/libraries/Builder/Microsoft.Agents.Builder.Dialogs/Prompts/OAuthPrompt.cs
```diff
@@ -179,69 +179,75 @@ public override async Task<DialogTurnResult> ContinueDialogAsync(DialogContext d
...
-                        var serviceUrl = dc.Context.Activity.ServiceUrl;
-                        var audience = callerInfo.Scope;
-                        var connectorClient = await CreateConnectorClientAsync(dc.Context, audience, cancellationToken).ConfigureAwait(false);
-                        dc.Context.Services.Set<IConnectorClient>(connectorClient);
-                    }
-                }
+            if (flowResult.Status == OAuthFlowStatus.UserCancelled)
+            {
+                // Preserve pre-existing behavior: surface failure by propagating UserCancelledException.
+                throw new UserCancelledException();
```
tracyboehrer, line 194: Is a throw correct here?  Would not this be a similar handling as Timeout (above this)?

src/libraries/Builder/Microsoft.Agents.Builder/UserAuth/TokenService/OAuthFlow.cs
tracyboehrer: I wonder... Should this be an exception?  This is Teams telling us it failed (which is usually because of a config error).  Maybe an exception is correct.  Worth pondering more.

We could say... OAuthFlow only ever returns the new OAuthFlowResult.  Even for BeginTurn.

It would only every throw for exceptional reasons.  OAuthFlowResult would then be: Pending, Completed with token, Completed with error.  And throw when it all went sideways.

AzureBotUserAuthorization would then translate that to a TokenResponse result, or throw.  Which is weird.  If I could do all this again, I would use that `SignInResponse` throughout.  But let's not go that far this round.

src/libraries/Builder/Microsoft.Agents.Builder.Dialogs/Prompts/OAuthPrompt.cs
sw-joelmut: This was an exception that oAuthFlow threw, so i moved it here to preserve this behavior and not break oAuthPrompt, since the try/catch we had here only capture Timeout, and the rest was bubble up to this function caller.

src/libraries/Builder/Microsoft.Agents.Builder/UserAuth/TokenService/OAuthFlow.cs
sw-joelmut: Changes applied!

Regarding SinginResponse, yea, we could use that, also i see the AuthExceptionReason has some of these values as well, so maybe at some point we can consolidate all these?

src/libraries/Builder/Microsoft.Agents.Builder.Dialogs/Prompts/OAuthPrompt.cs
tracyboehrer: I suspect that the dialog state will be wrong and that we should just end the dialog.  i.e., It could have been a latent bug that it was throwing that?

src/libraries/Builder/Microsoft.Agents.Builder/UserAuth/TokenService/OAuthFlow.cs
tracyboehrer: Long term it would be best to carry this all the way up, but that starts changing public surface.  Though, highly unlikely anyone directly uses or implements those.

tracyboehrer (approved):

Issue #842, opened 2026-05-25
DictionaryOfObjectConverter.SerializeJsonArray throws InvalidOperationException when array contains primitive-type elements


mdesalvo:
## Version
1.5.184

**Describe the bug**
`DictionaryOfObjectConverter.Write` crashes with **System.InvalidOperationException** at `JsonNode.AsObject()` when a Dictionary<string, object> value contains an array of primitive types (e.g., string[], int[], bool[]).

The root cause is in `SerializeJsonArray` (Serialization/Converters/DictionaryOfObjectConverter.cs, lines 220-231): it unconditionally calls `AddTypeInfo()` on every array element, which in turn calls `jsonNode.AsObject()` (Serialization/TypeExtensions.cs, line 20). This assumes every element serializes to a JsonObject, but primitive-type elements serialize to **JsonValue** nodes — and `JsonValue.AsObject()` throws InvalidOperationException.

**To Reproduce**
Store a string[] (or any primitive array) in a Dictionary<string, object> that is serialized by DictionaryOfObjectConverter:

```csharp
// In an Agents SDK dialog:
stepContext.Values.Add("MyKey", new string[] { "foo", "bar", "baz" });

// Then end the turn with a Prompt (triggers state serialization to storage):
return await stepContext.PromptAsync(nameof(TextPrompt), promptOptions, cancellationToken);
// --> InvalidOperationException at state save
```

**Expected behavior**
Serialization should succeed. `SerializeJsonArray` should handle array elements that are not JsonObject (i.e., JsonValue or nested JsonArray).

**Actual Behavior**
```csharp
System.InvalidOperationException:
 at System.Text.Json.ThrowHelper.ThrowInvalidOperationException_NodeWrongType(...)
 at System.Text.Json.Nodes.JsonNode.AsObject(...)
 at Microsoft.Agents.Core.Serialization.TypeExtensions.AddTypeInfo(JsonNode jsonNode, Object value)
 at Microsoft.Agents.Core.Serialization.Converters.DictionaryOfObjectConverter.SerializeJsonArray(JsonArray jArray, IList sourceList)
 at Microsoft.Agents.Core.Serialization.Converters.DictionaryOfObjectConverter.Write(...)
```

**Hosting Information:**
 - How are you Hosting this: **Azure**
 - Are you deploying: **AppServices** 
 - Are you using Azure Bot Services: **Yes**
 - What Client are you using: **other**
 - What .net version is your build in: **8**

**Additional context**
[...]

PR #843, opened 2026-05-26
DictionaryOfObjectConverter.SerializeJsonArray to proper handle primitive-type elements

mdesalvo:
This PR addresses #842 in order to enhance `DictionaryOfObjectConverter` for covering the serialization/deserialization flow of `IDictionary<string, object>` when it contains **arrays of primitives** (e.g: bool[], int[], string[]). In detail, it contains the following interventions:

### Bug Fixes

1. Serialization — `SerializeJsonArray`crashes on arrays of primitives
`AsObject()` called on each array element without checking its type. Primitives (string, int, bool) serialize as `JsonValue`, not `JsonObject` → InvalidOperationException.
Fix: Three-branch pattern match (`JsonObject` / `JsonArray` / skip for `JsonValue`), with support for recursion on nested arrays.

2. Serialization — `Write`crashes on `JsonElement` arrays
The explicit cast `(IList)item.Value` fails when the value is a `JsonElement` of kind `Array` (common with STJ when the target type is object) → InvalidCastException.
Fix: safe-check `item.Value` is `IList` in the condition.

3. Deserialization — `DeserializeJsonArray`→ NullReferenceException
`objValue` initialized only within the branch with `$type` metadata. For primitive arrays (without metadata), it remains null → crash on `objValue.Add()`.
Fix: Initialize `List<object>` to the first element without type info, and call `DeserializeJsonValue()` to extract CLR values ​​instead of adding raw JsonNodes.

4. TypeExtensions — 3 methods with unguarded `AsObject()`
- `AddTypeInfo(JsonNode)` — serialization
- `GetTypeInfo(JsonNode)` — deserialization
- `RemoveTypeInfo(JsonNode)` — deserialization
All were called `jsonNode.AsObject()` without guards.
Fix: pattern match `if (jsonNode is JsonObject)`.

### Real-world scenario

`WaterfallDialog` saves `DialogInstance.State` as `IDictionary<string, object>`. When a multi-turn dialog receives parameters as **string[]** (in `state["options"]`) or stores them in `stepContext.Values`, saving the state (e.g., to Azure Blob Storage) triggers `DictionaryOfObjectConverter.Write` → crash. This renders any `ComponentDialog` that passes arrays of primitives as options or intermediate values unusable!

### Tests added

14 tests in `DictionaryOfObjectConverterTests.cs`:
- serialization (string[], int[], bool[], JsonElement array, mixed content, empty array)
- WaterfallDialog (options, values, both)
- roundtrip with value verification,
- complex object array compatibility
All tests green on net8.0 and net4.8

### DIFF
--- a/src/libraries/Core/Microsoft.Agents.Core/Serialization/Converters/DictionaryOfObjectConverter.cs
+++ b/src/libraries/Core/Microsoft.Agents.Core/Serialization/Converters/DictionaryOfObjectConverter.cs
@@ -140,9 +140,9 @@ namespace Microsoft.Agents.Core.Serialization.Converters
                     {
                         SerializeJsonObject(jObj, item.Value, options);
                     }
-                    else if (newValue is JsonArray jArray)
+                    else if (newValue is JsonArray jArray && item.Value is IList sourceList)
                     {
-                        SerializeJsonArray(jArray, (IList)item.Value);
+                        SerializeJsonArray(jArray, sourceList);
                     }
                 }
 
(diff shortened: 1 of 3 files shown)


mdesalvo:

tracyboehrer (approved):

tracyboehrer:
Thanks for the submission!

PR #883, opened 2026-06-19
Extend IStorage interface to support new operations

### LINKED ISSUE #848: Extend IStorage interface to support new operations
## Description

We should have an extended interface for IStorage with the new operations (WriteAsync overload).
There is possibly another operation about getting a paginated list of records, assuming we can abstract that across providers.

More information in this PR https://github.com/microsoft/Agents-for-net/pull/714#issuecomment-4565674441

sw-joelmut:
Fixes #848

**Description**
This PR extends the storage system by introducing IStorageV2 interface to support new operations with structured result types. It adds a StorageCompatibility helper to maintain backward compatibility while enabling enhanced storage capabilities.

**Key changes**
- New storage interface: Added IStorageV2 interface that extends IStorage with new methods returning structured results (StorageOperationStatus, typed values)
- Compatibility layer: Introduced StorageCompatibility class to convert between IStorage and IStorageV2, ensuring existing code continues to work
- Storage implementations: Updated MemoryStorage, BlobsStorage, and CosmosDbPartitionedStorage to implement IStorageV2
- Component updates: Modified AgentState, UserAuthorization, AzureBotUserAuthorization, ConfigurationAgentHost, and other components to use IStorageV2 and handle structured storage results
- Enhanced operations: Storage operations now return status codes (Succeeded, NotFound, Conflict) and typed values instead of nullable objects
- Test coverage: Updated existing tests and added new tests for IStorageV2 functionality

**Testing**
The following image shows a few of the tested scenarios.
image


MattB-msft (dismissed):

sw-joelmut:
Included fix for issue #1012

Reproduction:
image

Fix:
image

NickSpag:
@sw-joelmut found a regression in 480118d9: a turn that calls `DeleteStateAsync` now throws `EtagException` from the end-of-turn save, even when no other turn touched the conversation.

`DeleteStateAsync` deletes the document and calls `ClearState`, which sets `Hash = string.Empty` to force a save ([AgentState.cs#L498](https://github.com/microsoft/Agents-for-net/blob/480118d9d897785610386278b12acf7fb9872486/src/libraries/Builder/Microsoft.Agents.Builder/State/AgentState.cs#L498)), but it doesn't reset `cachedState.Version`. The end-of-turn save then passes the version loaded before the delete as `ExpectedVersion` ([#L266](https://github.com/microsoft/Agents-for-net/blob/480118d9d897785610386278b12acf7fb9872486/src/libraries/Builder/Microsoft.Agents.Builder/State/AgentState.cs#L266)). The document no longer exists, so storage returns `ConditionNotMet` and the save throws "another turn updated the state first" ([#L277](https://github.com/microsoft/Agents-for-net/blob/480118d9d897785610386278b12acf7fb9872486/src/libraries/Builder/Microsoft.Agents.Builder/State/AgentState.cs#L277)). Before this commit, the same save wrote an empty document back.

Repro against `MemoryStorage`:

```csharp
var storage = new MemoryStorage();
var seed = new ConversationState(storage);
var seedContext = TestUtilities.CreateEmptyContext();
await seed.LoadAsync(seedContext);
seed.SetValue("k", "v");
await seed.SaveChangesAsync(seedContext);

var state = new ConversationState(storage);
var context = TestUtilities.CreateEmptyContext();
await state.LoadAsync(context);
await state.DeleteStateAsync(context);
await state.SaveChangesAsync(context); // throws EtagException (ConditionNotMet)
```

The same sequence with `ClearState()` in place of `DeleteStateAsync` passes.

The repo's samples hit this: the EndOfConversation handler in `CopilotStudioEchoSkill/EchoSkill.cs` and both handlers in `AgentToAgent/Agent2/Echo.cs` delete conversation state in a normal route. The call in `AgentResponses.cs` is unaffected, because it runs in `OnTurnError`, which rethrows before the save.

a potential fix: in `DeleteStateAsync`, clear the cached version and leave the cleared cache marked unchanged instead of forcing a save. The end-of-turn save then does nothing unless the handler sets new values after the delete. `ClearState` on its own keeps its forced save.

The repro only ran against MemoryStorage. BlobsStorage has the same version check, so it should fail the same way, but I didn't run it.

sw-joelmut:
> Joel Mut (@sw-joelmut) found a regression in [480118d](https://github.com/microsoft/Agents-for-net/commit/480118d9d897785610386278b12acf7fb9872486): a turn that calls `DeleteStateAsync` now throws `EtagException` from the end-of-turn save, even when no other turn touched the conversation.
> 
> `DeleteStateAsync` deletes the document and calls `ClearState`, which sets `Hash = string.Empty` to force a save ([AgentState.cs#L498](https://github.com/microsoft/Agents-for-net/blob/480118d9d897785610386278b12acf7fb9872486/src/libraries/Builder/Microsoft.Agents.Builder/State/AgentState.cs#L498)), but it doesn't reset `cachedState.Version`. The end-of-turn save then passes the version loaded before the delete as `ExpectedVersion` ([#L266](https://github.com/microsoft/Agents-for-net/blob/480118d9d897785610386278b12acf7fb9872486/src/libraries/Builder/Microsoft.Agents.Builder/State/AgentState.cs#L266)). The document no longer exists, so storage returns `ConditionNotMet` and the save throws "another turn updated the state first" ([#L277](https://github.com/microsoft/Agents-for-net/blob/480118d9d897785610386278b12acf7fb9872486/src/libraries/Builder/Microsoft.Agents.Builder/State/AgentState.cs#L277)). Before this commit, the same save wrote an empty document back.
> 
> Repro against `MemoryStorage`:
> 
> ```cs
> var storage = new MemoryStorage();
> var seed = new ConversationState(storage);
> var seedContext = TestUtilities.CreateEmptyContext();
> await seed.LoadAsync(seedContext);
> seed.SetValue("k", "v");
> await seed.SaveChangesAsync(seedContext);
> 
> var state = new ConversationState(storage);
> var context = TestUtilities.CreateEmptyContext();
> await state.LoadAsync(context);
> await state.DeleteStateAsync(context);
> await state.SaveChangesAsync(context); // throws EtagException (ConditionNotMet)
> ```
> 
> The same sequence with `ClearState()` in place of `DeleteStateAsync` passes.
> 
> The repo's samples hit this: the EndOfConversation handler in `CopilotStudioEchoSkill/EchoSkill.cs` and both handlers in `AgentToAgent/Agent2/Echo.cs` delete conversation state in a normal route. The call in `AgentResponses.cs` is unaffected, because it runs in `OnTurnError`, which rethrows before the save.
> 
> a potential fix: in `DeleteStateAsync`, clear the cached version and leave the cleared cache marked unchanged instead of forcing a save. The end-of-turn save then does nothing unless the handler sets new values after the delete. `ClearState` on its own keeps its forced save.
> 
> The repro only ran against MemoryStorage. BlobsStorage has the same version check, so it should fail the same way, but I didn't run it.

Thanks Nick, great catch. I reproduced it and confirmed that `DeleteStateAsync` leaves the loaded version behind, causing the end-of-turn save to fail.
I added a regression test and updated the delete path to clear that cached version and prevent the automatic save from recreating the deleted state.

Also, tested this fix with Memory, Blobs, and Cosmos, all of them are working.
image

NickSpag:
awesome, glad it helped!

tracyboehrer (approved):

PR #945, opened 2026-07-29
Multi-adapter registry (IChannelAdapterRegistry)

tracyboehrer:
## Purpose

Enable a single agent to host **multiple protocol adapters** (Activity Protocol, A2A, and
future protocols) and resolve the correct one by `channelId`, instead of being limited to
one DI-registered `IChannelAdapter`.

## Key points

- **`IChannelAdapterRegistry`** (Builder layer), surfaced on
  `AgentApplicationOptions.ChannelAdapterRegistry`, lets code look up an adapter by
  `channelId` rather than manually wiring each one into DI.
- Adapters opt in with a class-level **`[ChannelAdapter("channelId")]`** attribute. A Roslyn
  source generator (shipped in `Microsoft.Agents.Core`) enrolls them automatically — no
  manual registration call.
- **Agent Extensions can define and ship their own adapters** this way. For example,
  `A2AAdapter` is annotated `[ChannelAdapter("a2a")]` and is discovered with no extra wiring.
- **Two-tier HTTP dispatch:** Tier 1 = protocol adapters on their own dedicated endpoints;
  Tier 2 = the shared `/api/messages` peeks the inbound Activity's `channelId` and routes to
  a channel-specific Activity Protocol adapter, falling back to CloudAdapter.
- Proactive messaging gains `channelId`/`Conversation` overloads that resolve the adapter via
  the registry.
- **Additive / non-breaking.** Existing single-adapter agents are unaffected — the Tier 2
  peek only runs when channel-specific adapters are registered.

Full design details are in the design doc attached as a comment on this PR.


tracyboehrer:
**Design doc:** `docs/multi-protocol/2026-06-15-multi-adapter-registry-addendum.md` (attached below for full details).

<details>
<summary>Multi-Adapter Registry &amp; Dispatch Design — full design doc</summary>

# Multi-Adapter Registry & Dispatch Design

## Addendum to: Non-Breaking Multi-Adapter Extension Design (2026-06-10)

### Status: Implemented

This document describes the **Channel Adapter Registry** and **multi-adapter endpoint
dispatch** mechanism that enables multiple protocol adapters to coexist within a single
agent application. It is a child of the [Non-Breaking Multi-Protocol Design](2026-06-10-non-breaking-multi-protocol-addendum.md)
for clarity — that document covers the extension pattern and custom context design;
this one covers the infrastructure that enables multi-adapter routing.

> **Implementation note.** The shipped design differs from the original proposal in one
> important way: **the registry works in terms of `IChannelAdapter`, not
> `IAgentHttpAdapter`.** The interface lives in the Builder layer (which cannot see
> `IAgentHttpAdapter`), so it is surfaced on `AgentApplicationOptions.ChannelAdapterRegistry`
> and returns `IChannelAdapter`. Tier 2 HTTP dispatch casts the resolved adapter to
> `IAgentHttpAdapter`.

---

## Motivation

Today, `AgentApplication` resolves a single `IChannelAdapter` (CloudAdapter) from DI —
there is exactly one, and every inbound request goes through it. To support multiple
protocols (A2A, MCP, future non-AP protocols) alongside the Activity Protocol — each with
its own adapter — and to make working with more than one adapter ergonomic, we need:

1. A **registry** that maps channel IDs to their adapter implementations. Without it,
   developers must manually register each additional adapter in DI and hold a reference to
   the right one wherever they need it. The registry lets them simply look an adapter up by
   channelId.
2. An **endpoint mapping** mechanism that routes HTTP requests to the correct adapter
3. A **dispatch model** that handles both dedicated endpoints (e.g., `/a2a`) and shared
   endpoints (e.g., `/api/messages`) serving multiple AP channels

---

## Channel Adapter Registry

The core framework addition enabling multi-adapter support. The interface lives in the
**Builder layer** (`Microsoft.Agents.Builder.App`, multi-target). The implementation lives
in **Hosting.AspNetCore** (`net8.0`).

Because the Builder layer cannot reference `IAgentHttpAdapter` (that type is defined in
Hosting.AspNetCore, which references Builder — not the other way around), the registry is
expressed in terms of the Builder-native `IChannelAdapter`. Both real adapters
(`CloudAdapter` and `A2AAdapter`) implement **both** `IChannelAdapter` and
`IAgentHttpAdapter`, so Tier 2 HTTP dispatch can cast a resolved adapter to the HTTP
contract when it needs to serve a request.

```csharp
namespace Microsoft.Agents.Builder.App
{
    /// <summary>
    /// Registry of <see cref="IChannelAdapter"/> instances keyed by channelId.
    /// Lets SDK features and developers resolve the correct adapter for a given channel
    /// without needing to know adapter types. Surfaced on
    /// <see cref="AgentApplicationOptions.ChannelAdapterRegistry"/>.
    /// </summary>
    public interface IChannelAdapterRegistry
    {
        /// <summary>
        /// Gets the adapter registered for the specified channelId. Throws if not found.
        /// </summary>
        IChannelAdapter GetAdapter(string channelId);

        /// <summary>
        /// Gets the default Activity Protocol adapter (CloudAdapter). Used as fallback
        /// when no channel-specific adapter is registered.
        /// </summary>
        IChannelAdapter GetDefault();

        /// <summary>
        /// Tries to get the adapter for the specified channelId. Returns false (and a
        /// null adapter) if no adapter is registered for that channel.
        /// </summary>
        bool TryGetAdapter(string channelId, out IChannelAdapter adapter);

        /// <summary>
        /// Whether any channel-specific adapters are registered beyond the default.
        /// When false, the Tier 2 channelId peek in MapAgentApplicationEndpoints is
        /// skipped entirely — preserving current performance.
        /// </summary>
        bool HasChannelSpecificAdapters { get; }

        /// <summary>
        /// Gets all registered adapters (default + channel-specific).
        /// </summary>
        IEnumerable<IChannelAdapter> GetAll();
    }
}
```

The registry is populated at construction from two sources (see
[Registry Population](#registry-population)):

1. Attribute-discovered registrations emitted by the source generator, and
2. Explicit `ChannelAdapterRegistration` descriptors registered in DI (which override
   discovered ones for the same channelId).

The existing `IChannelAdapter` registration (CloudAdapter) becomes the **default adapter**.

### Exposed on `AgentApplicationOptions`

Developers (and SDK features) resolve adapters through the options object:

```csharp
public class AgentApplicationOptions
{
    // ...
    /// <summary>
    /// Registry of adapters keyed by channelId. Populated from DI when available; may be
    /// null when options are constructed programmatically (e.g., in unit tests).
    /// </summary>
    public IChannelAdapterRegistry? ChannelAdapterRegistry { get; set; }
}
```

The property is populated in the DI constructor via
`sp.GetService<IChannelAdapterRegistry>()`.

---

## IChannelAdapter — No Changes Required

The existing `IChannelAdapter` interface is **unchanged**. Adapters are keyed by the
`channelId` string supplied to the `[ChannelAdapter("channelId")]` attribute, not by a
property on the adapter itself. This avoids adding a breaking member to `IChannelAdapter`
and keeps adapter implementations simple.

---

## Adapter Registration — `[ChannelAdapter]` Attribute Pipeline

Rather than a DI registration method, an adapter opts into the registry with a class-level
attribute. This mirrors the serialization initialization pattern
(`ActivityTypeAttribute` → source generator → `ActivityTypeInitAssemblyAttribute` →
runtime discovery).

### 1. Class-level attribute (authored by the adapter)

```csharp
// Microsoft.Agents.Hosting.AspNetCore
[AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = false)]
public sealed class ChannelAdapterAttribute : Attribute
{
    public ChannelAdapterAttribute(string channelId) => ChannelId = channelId;

    /// <summary>The channelId this adapter handles (e.g., "msteams", "a2a").</summary>
    public string ChannelId { get; }
}
```

Applied to an adapter that implements `IChannelAdapter` (and, to serve HTTP traffic,
`IAgentHttpAdapter`):

```csharp
[ChannelAdapter("a2a")]
public class A2AAdapter : ChannelAdapter, IAgentHttpAdapter, IA2AHttpAdapter { ... }
```

### 2. Source generator (emits the assembly attribute)

`ChannelAdapterInitSourceGenerator` lives in **`Microsoft.Agents.Core.Analyzers`**
(a netstandard2.0 Roslyn generator). It matches every type annotated with
`Microsoft.Agents.Hosting.AspNetCore.ChannelAdapterAttribute` via
`ForAttributeWithMetadataName` and emits one assembly-level attribute per annotation into
`ChannelAdapterInitAssemblyAttributes.g.cs`:

```csharp
[assembly: Microsoft.Agents.Hosting.AspNetCore.ChannelAdapterInitAssemblyAttribute(
    typeof(MyNamespace.A2AAdapter), "a2a")]
```

**Packaging.** The generator ships **inside the `Microsoft.Agents.Core` NuGet package** as
an analyzer, so NuGet consumers get it automatically. In-solution project references do not
run packaged analyzers, so projects that define annotated adapters (e.g.
`Microsoft.Agents.Hosting.AspNetCore.A2A`) add an explicit analyzer reference:

```xml
<ProjectReference Include="..\..\Core\Microsoft.Agents.Core.Analyzers\Microsoft.Agents.Core.Analyzers.csproj"
                  OutputItemType="Analyzer"
                  ReferenceOutputAssembly="false" />
```

### 3. Assembly attribute (discovery surface)

```csharp
// Microsoft.Agents.Hosting.AspNetCore
[AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
public sealed class ChannelAdapterInitAssemblyAttribute : Attribute
{
    /// <param name="adapterType">The annotated adapter type. Must implement IChannelAdapter.</param>
    /// <param name="channelId">The channelId the adapter handles.</param>
    public ChannelAdapterInitAssemblyAttribute(Type adapterType, string channelId) { ... }

    /// <summary>
    /// Reads ChannelAdapterInitAssemblyAttribute instances off the loaded assemblies and
    /// projects them to ChannelAdapterRegistration descriptors. Filters to types that
    /// implement IChannelAdapter.
    /// </summary>
    public static IEnumerable<ChannelAdapterRegistration> GetRegistrations() { ... }
}
```

At registry construction, `GetRegistrations()` scans loaded assemblies for these
attributes — no runtime type scanning of every type is required, and adapter authors never
call an explicit DI registration method.

### Optional: explicit DI descriptors

For advanced scenarios (or to override a discovered registration), a
`ChannelAdapterRegistration` can be registered directly in DI. DI descriptors take
precedence over attribute-discovered ones for the same channelId.

```csharp
public sealed record ChannelAdapterRegistration(string ChannelId, Type AdapterType);

// e.g., override the "msteams" adapter:
services.AddSingleton(new ChannelAdapterRegistration("msteams", typeof(MyTeamsAdapter)));
```

---

## Adapter Selection — Two-Tier Model

Adapter selection uses a **two-tier** model that handles both protocol-level separation
and channel-level specialization within a shared endpoint:

1. **Endpoint-based (Tier 1):** The HTTP route forces a specific adapter. This handles
   fundamentally different wire protocols that live on their **own dedicated endpoints** —
   i.e., *not* `/api/messages`. For example, a hypothetical A2A adapter on `/a2a` or a
   hypothetical Discord adapter on `/discord`. No payload inspection needed — the endpoint
   *is* the selector.

2. **ChannelId-based (Tier 2):** For the **shared** `/api/messages` endpoint, where
   Azure Bot Service routes multiple Activity Protocol channels (Teams, Slack, WebChat,
   DirectLine) through a single messaging endpoint, the handler peeks at the `channelId` in
   the inbound Activity and resolves a channel-specific adapter from the registry. These are
   still Activity Protocol adapters — e.g., a hypothetical Teams or Slack adapter that lives
   on `/api/messages` but needs its own adapter customizations. If no channel-specific
   adapter is registered (or it cannot serve HTTP), the default AP adapter (CloudAdapter)
   handles the request.

**Why two tiers are necessary:** Azure Bot Service configures a single messaging
endpoint on the Bot resource. All ABS channels — Teams, Slack, WebChat, DirectLine —
hit that same URL. If a custom `TeamsAdapter` or `SlackAdapter` is needed
(for channel-specific pipeline behavior, middleware, or context enrichment), endpoint-
based selection alone cannot disambiguate. The payload must be inspected.

### AgentInterfaceAttribute (Existing)

The attribute already exists in the codebase. Each Map method reads these attributes
and filters by its own protocol:

```csharp
// Already defined in Microsoft.Agents.Builder.App
[AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = true)]
public class AgentInterfaceAttribute(string protocol, string path, string? processDelegate = null) : Attribute
{
    public string Protocol { get; } = protocol;
    public string Path { get; } = path;
    public string? ProcessDelegate { get; } = processDelegate;
}
```

- `MapAgentApplicationEndpoints` filters for `Protocol == AgentTransportProtocol.ActivityProtocol`
- `MapA2AApplicationEndpoints` filters for `Protocol == A2AAgentTransportProtocol.JsonRpc` or `HttpJson`
- Each Map method skips attributes for other protocols

### Endpoint Mapping — MapAgentApplicationEndpoints with Tier 2

`MapAgentApplicationEndpoints` already reads `[AgentInterface]` attributes and maps
endpoints. The change is: when `HasChannelSpecificAdapters` is true, the handler
uses the registry for Tier 2 resolution instead of always using the single
`IAgentHttpAdapter`:

```csharp
// Updated handler inside MapAgentApplicationEndpoints
agentGroup.MapMethods(agentInterface.Path, ["POST"],
    async (HttpRequest request, HttpResponse response,
           IAgentHttpAdapter defaultAdapter,
           IChannelAdapterRegistry registry,
           IServiceProvider services,
           CancellationToken ct) =>
    {
        // Resolve adapter: Tier 2 if channel-specific adapters exist, else default
        var adapter = registry.HasChannelSpecificAdapters
            ? await ResolveAdapterAsync(registry, defaultAdapter, request, ct)
            : defaultAdapter;

        IAgent agentInstance = (IAgent)services.GetService(agent)
            ?? (IAgent)services.GetRequiredService(typeof(IAgent));

        if (!string.IsNullOrEmpty(agentInterface.ProcessDelegate))
        {
            // Custom process delegate (existing behavior)
            var processDelegate = /* resolve delegate */;
            await processDelegate(request, response, adapter, agentInstance, ct);
        }
        else
        {
            await adapter.ProcessAsync(request, response, agentInstance, ct);
        }
    });
```

### Adapter Resolution Logic (Tier 2 Only)

This code is only reached when channel-specific AP adapters are registered. Because the
registry returns `IChannelAdapter`, the resolved adapter is cast to `IAgentHttpAdapter` to
serve the HTTP endpoint; if the channel adapter cannot serve HTTP, dispatch falls back to
the default adapter:

```csharp
internal static async ValueTask<IAgentHttpAdapter> ResolveAdapterAsync(
    IChannelAdapterRegistry registry,
    IAgentHttpAdapter defaultAdapter,
    HttpRequest request,
    CancellationToken cancellationToken)
{
    var channelId = await PeekChannelIdAsync(request, cancellationToken).ConfigureAwait(false);

    // A channel-specific adapter must also handle HTTP (IAgentHttpAdapter) to serve this endpoint.
    if (channelId != null
        && registry.TryGetAdapter(channelId, out var channelAdapter)
        && channelAdapter is IAgentHttpAdapter httpAdapter)
    {
        return httpAdapter;
    }

    // Default: CloudAdapter handles all unspecialized AP channels.
    return defaultAdapter;
}
```

### PeekChannelIdAsync — Lightweight Payload Inspection

The `channelId` field is a top-level property in the Activity JSON, but JSON has no
guaranteed property ordering — it could appear anywhere in the payload. Reading it
does not require full *deserialization* into an Activity object, but does require
scanning the full top-level object. The request body is buffered (so the adapter can
re-read it), then a streaming `Utf8JsonReader` extracts the value without allocating model
objects:

```csharp
private static async ValueTask<string> PeekChannelIdAsync(
    HttpRequest request, CancellationToken cancellationToken)
{
    // Enable buffering so the adapter can re-read the full body in ProcessAsync.
    request.EnableBuffering();

    var body = request.Body;
    var contentLength = (int)(request.ContentLength ?? 0);

    // Read the full body: from a pooled buffer when ContentLength is known, or via a
    // MemoryStream fallback for chunked/unknown-length requests. Reset body.Position = 0
    // afterwards so the resolved adapter reads the intact payload.
    // ... (ArrayPool<byte> rent/return, or MemoryStream.CopyToAsync) ...

    return TryReadChannelId(buffer);   // see below
}

// Zero-allocation scan for the TOP-LEVEL channelId only.
internal static string TryReadChannelId(ReadOnlySpan<byte> utf8Json)
{
    try
    {
        var reader = new Utf8JsonReader(utf8Json);
        while (reader.Read())
        {
            if (reader.CurrentDepth == 1
                && reader.TokenType == JsonTokenType.PropertyName
                && reader.ValueTextEquals("channelId"u8))
            {
                return reader.Read() && reader.TokenType == JsonTokenType.String
                    ? reader.GetString()
                    : null;
            }
        }
    }
    catch (JsonException)
    {
        // Malformed/partial JSON — fall back to the default adapter.
    }

    return null;
}
```

**Behavior and performance:**
- Only the **top-level** `channelId` is honored (`CurrentDepth == 1`); a nested
  `from.channelId` decoy is ignored.
- Uses `Utf8JsonReader` (zero-allocation streaming reader) — no model objects allocated,
  and nested objects/arrays (e.g. large `attachments`/`entities`) are skipped by the reader.
- Malformed, empty, or missing-`channelId` bodies return `null` → dispatch falls back to
  the default adapter.
- `EnableBuffering()` + `body.Position = 0` leave the body re-readable so the resolved
  adapter re-parses the full Activity in `ProcessAsync` — an unavoidable Tier 2 cost, but
  negligible for typical payloads relative to turn execution time.
- For non-AP endpoints (Tier 1), and when `HasChannelSpecificAdapters` is false, this code
  is never reached — zero overhead.

### Channel-Specific Adapter Example

A `TeamsAdapter` registers alongside CloudAdapter and handles only Teams traffic
routed through `/api/messages`. It does NOT need to derive from CloudAdapter — it just
implements `IChannelAdapter` + `IAgentHttpAdapter` and carries the `[ChannelAdapter]`
attribute:

```csharp
// Option A: Derives from CloudAdapter (shares AP pipeline)
[ChannelAdapter("msteams")]
public class TeamsChannelAdapter : CloudAdapter
{
    // Override pipeline behavior for Teams-specific needs
}

// Option B: Custom non-CloudAdapter implementation
[ChannelAdapter("msteams")]
public class TeamsAdapter : ChannelAdapter, IAgentHttpAdapter
{
    public TeamsAdapter(IChannelServiceClientFactory clientFactory, ...) : base(logger) { }

    public Task ProcessAsync(HttpRequest httpRequest, HttpResponse httpResponse,
        IAgent agent, CancellationToken ct) { /* custom */ }
}
```

When a request arrives at `/api/messages` with `"channelId": "msteams"`, the
endpoint handler resolves the registered TeamsAdapter. All other channels (Slack,
WebChat, DirectLine) fall through to CloudAdapter.

### When Tier 2 Is Not Needed

If no channel-specific adapters are registered — which is the common case today — the
Tier 2 peek is bypassed entirely and the default adapter is used immediately. This
preserves current performance characteristics for the vast majority of agents. When no
`[AgentInterface]` attributes are present, the existing behavior applies —
`/api/messages` maps to the default AP adapter. This preserves backward compatibility.

---

## DI Registration

### Existing Behavior (Unchanged)

`AddAgent<T>()` continues to register CloudAdapter (or a CloudAdapter subclass) as
the default AP adapter. This is the only adapter most agents need:

```csharp
builder.AddAgent<MyAgent>();
// Internally:
//   services.AddSingleton<CloudAdapter>();
//   services.AddSingleton<IAgentHttpAdapter>(sp => sp.GetService<CloudAdapter>());
//   services.AddSingleton<IChannelAdapter>(sp => sp.GetService<CloudAdapter>());
```

### How Adapters Enter the Registry

| Vector | Who uses it | Registry key | Precedence |
|--------|-------------|--------------|------------|
| `IChannelAdapter` DI registration (CloudAdapter) | `AddAgent<T>()` | `(default)` — via `GetDefault()` | — |
| `[ChannelAdapter("channelId")]` attribute + source generator | Adapter author / extension | e.g., `"a2a"`, `"msteams"` | Discovered |
| `ChannelAdapterRegistration` descriptor in DI | Advanced / override | e.g., `"msteams"` | Overrides discovered for same channelId |

The default adapter is resolved lazily from DI by `GetDefault()` (`IChannelAdapter` →
CloudAdapter). Channel-specific adapters come from the attribute pipeline, optionally
overridden by DI descriptors.

### Extension Self-Annotation Pattern

Extensions (A2A, MCP, etc.) annotate their adapter and provide their own `AddXxxAdapter()`
for the protocol-specific interface + `MapXxxEndpoints()` for their dedicated endpoint. The
`[ChannelAdapter]` attribute (via the generator) is what enrolls the adapter into the
registry — the `AddXxxAdapter()` method does not need to touch the registry:

```csharp
// Inside Microsoft.Agents.Hosting.AspNetCore.A2A
[ChannelAdapter("a2a")]
public class A2AAdapter : ChannelAdapter, IAgentHttpAdapter, IA2AHttpAdapter { ... }

public static void AddA2AAdapter(this IServiceCollection services)
{
    services.AddSingleton<A2AAdapter>();
    services.AddSingleton<IA2AHttpAdapter>(sp => sp.GetService<A2AAdapter>());
    // No explicit registry call: the [ChannelAdapter("a2a")] attribute + generated
    // assembly attribute enroll A2AAdapter in the registry.
}
```

`MapA2AApplicationEndpoints()` resolves the adapter directly via `IA2AHttpAdapter`. The
registry registration enables SDK features to find the adapter by channelId but is NOT used
for that endpoint's resolution.

### Registry Population

`ChannelAdapterRegistry` merges the two sources at construction — attribute-discovered
registrations first, then DI descriptors (which win for the same channelId). Adapter
instances are resolved lazily via `ActivatorUtilities.GetServiceOrCreateInstance` (DI
singleton if registered, otherwise created) and cached, so annotated adapters need no
explicit DI registration:

```csharp
internal sealed class ChannelAdapterRegistry : IChannelAdapterRegistry
{
    private readonly IServiceProvider _services;
    private readonly Dictionary<string, Type> _channelAdapters;
    private readonly ConcurrentDictionary<Type, IChannelAdapter> _instances = new();
    private IChannelAdapter _default;

    public ChannelAdapterRegistry(
        IServiceProvider services,
        IEnumerable<ChannelAdapterRegistration> registrations)
    {
        _services = services;
        _channelAdapters = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);

        // Attribute-discovered registrations first...
        foreach (var r in ChannelAdapterInitAssemblyAttribute.GetRegistrations())
            _channelAdapters[r.ChannelId] = r.AdapterType;

        // ...then explicit DI registrations, which override discovered ones per channelId.
        foreach (var r in registrations ?? Enumerable.Empty<ChannelAdapterRegistration>())
            if (r?.ChannelId is { Length: > 0 } && r.AdapterType != null)
                _channelAdapters[r.ChannelId] = r.AdapterType;
    }

    public bool HasChannelSpecificAdapters => _channelAdapters.Count > 0;

    public IChannelAdapter GetDefault()
        => _default ??= _services.GetRequiredService<IChannelAdapter>();

    public IChannelAdapter GetAdapter(string channelId)
        => channelId != null && _channelAdapters.TryGetValue(channelId, out var type)
            ? Resolve(type)
            : throw new InvalidOperationException($"No adapter registered for channel '{channelId}'.");

    public bool TryGetAdapter(string channelId, out IChannelAdapter adapter)
    {
        if (channelId != null && _channelAdapters.TryGetValue(channelId, out var type))
        {
            adapter = Resolve(type);
            return true;
        }
        adapter = null;
        return false;
    }

    public IEnumerable<IChannelAdapter> GetAll()
    {
        yield return GetDefault();
        foreach (var type in _channelAdapters.Values)
            yield return Resolve(type);
    }

    private IChannelAdapter Resolve(Type type)
        => _instances.GetOrAdd(type,
            t => (IChannelAdapter)ActivatorUtilities.GetServiceOrCreateInstance(_services, t));
}
```

The registry singleton is registered during `AddCloudAdapter<T>()`:

```csharp
services.TryAddSingleton<IChannelAdapterRegistry, ChannelAdapterRegistry>();
```

---

## Scenario Walkthroughs

### Scenario 1: MyAgent + A2A + Teams via default CloudAdapter

The simplest multi-protocol case. One agent handles Activity Protocol (via
CloudAdapter) and A2A (via A2AAdapter). No custom AP adapter needed.

```csharp
[AgentInterface(AgentTransportProtocol.ActivityProtocol, "/api/messages")]
[AgentInterface(A2AAgentTransportProtocol.JsonRpc, "/a2a")]
class MyAgent : AgentApplication { ... }

// Program.cs
builder.AddAgent<MyAgent>();          // CloudAdapter as default → registry GetDefault()
builder.Services.AddA2AAdapter();     // A2AAdapter → IA2AHttpAdapter; [ChannelAdapter("a2a")] enrolls it

app.MapAgentApplicationEndpoints();   // Reads [AgentInterface], maps /api/messages → IAgentHttpAdapter
app.MapA2AApplicationEndpoints();     // Reads [AgentInterface], maps /a2a → IA2AHttpAdapter
```

**Endpoint resolution:**
- `MapAgentApplicationEndpoints` resolves `IAgentHttpAdapter` from DI → CloudAdapter.
  Tier 2 is still active because `HasChannelSpecificAdapters` is true (A2A is registered),
  but for AP channels the peek finds no channel-specific AP adapter → default CloudAdapter.
- `MapA2AApplicationEndpoints` resolves `IA2AHttpAdapter` from DI → A2AAdapter.

**Adapter lookup (via `options.ChannelAdapterRegistry`):**
- `channelId: "a2a"` → registry → A2AAdapter ✓
- `channelId: "msteams"` → no specific adapter → `GetDefault()` → CloudAdapter ✓

### Scenario 2: MyAgent + A2A + custom TeamsAdapter (non-CloudAdapter)

The advanced case. TeamsAdapter handles Teams-specific traffic on the shared AP
endpoint. It does NOT derive from CloudAdapter.

```csharp
[AgentInterface(AgentTransportProtocol.ActivityProtocol, "/api/messages")]
[AgentInterface(A2AAgentTransportProtocol.JsonRpc, "/a2a")]
class MyAgent : AgentApplication { ... }

// TeamsAdapter — custom adapter for the Teams channel
[ChannelAdapter("msteams")]
public class TeamsAdapter : ChannelAdapter, IAgentHttpAdapter
{
    public TeamsAdapter(
        IChannelServiceClientFactory clientFactory,
        IActivityTaskQueue queue,
        ILogger<TeamsAdapter> logger) : base(logger)
    {
        // Teams-specific pipeline: custom middleware, context enrichment
    }

    public async Task ProcessAsync(HttpRequest httpRequest, HttpResponse httpResponse,
        IAgent agent, CancellationToken ct)
    {
        // Custom Teams processing
    }
}

// Program.cs — no AddChannelAdapter call needed; the attribute enrolls TeamsAdapter
builder.AddAgent<MyAgent>();          // CloudAdapter as default
builder.Services.AddA2AAdapter();     // A2A protocol

app.MapAgentApplicationEndpoints();   // /api/messages → Tier 2 resolution
app.MapA2AApplicationEndpoints();     // /a2a → IA2AHttpAdapter (A2AAdapter directly)
```

> If `TeamsAdapter` lives in the consuming app's own project, ensure the
> `Microsoft.Agents.Core` analyzer runs there (automatic for NuGet consumers; in-solution
> projects add the `OutputItemType="Analyzer"` reference shown earlier).

**Endpoint resolution at `/api/messages`:**
1. `MapAgentApplicationEndpoints` handler resolves `IChannelAdapterRegistry`.
2. `registry.HasChannelSpecificAdapters == true` → peeks `channelId` from body.
3. `"channelId": "msteams"` → `TryGetAdapter("msteams")` → TeamsAdapter, cast to
   `IAgentHttpAdapter` → TeamsAdapter serves the request.
4. `"channelId": "webchat"` → not found → default → CloudAdapter.

**Adapter lookup (via `options.ChannelAdapterRegistry`):**
- `channelId: "msteams"` → TeamsAdapter ✓
- `channelId: "directline"` → default → CloudAdapter ✓
- `channelId: "a2a"` → A2AAdapter ✓

### Why the attribute does NOT constrain to CloudAdapter

- `AddAgent<TAgent, TAdapter>()` registers the adapter as THE default AP adapter
  (replacing CloudAdapter in DI). Its `TAdapter : CloudAdapter` constraint ensures pipeline
  compatibility (auth, middleware, background processing).
- A `[ChannelAdapter("channelId")]`-annotated adapter registers a SUPPLEMENTAL adapter for
  a specific channel. It augments CloudAdapter rather than replacing it, and only needs to
  implement `IChannelAdapter` (plus `IAgentHttpAdapter` to serve HTTP traffic).

So: CloudAdapter subclasses that should be the default → `AddAgent<TAgent, TAdapter>()`;
channel-specific adapters → `[ChannelAdapter("channelId")]`.

---

## IAgentHttpAdapter — Unchanged Interface, Multiple Registrations

`IAgentHttpAdapter` remains unchanged. Multiple adapters implement it (CloudAdapter,
A2AAdapter, TeamsAdapter, etc.). Each is registered in DI under its own concrete type
and/or a protocol-specific interface.

**DI resolution remains the primary mechanism for Map methods:**
- `MapAgentApplicationEndpoints` resolves `IAgentHttpAdapter` → CloudAdapter (default)
- `MapA2AApplicationEndpoints` resolves `IA2AHttpAdapter` → A2AAdapter
- Tier 2: `MapAgentApplicationEndpoints` uses the registry only when
  `HasChannelSpecificAdapters == true`, casting the resolved `IChannelAdapter` to
  `IAgentHttpAdapter`.

**The registry is a lookup service, not a replacement for DI:**
```
DI registrations:
  IChannelAdapter   → CloudAdapter (default; registry GetDefault())
  IAgentHttpAdapter → CloudAdapter (resolved by MapAgentApplicationEndpoints)
  IA2AHttpAdapter   → A2AAdapter   (resolved by MapA2AApplicationEndpoints)

Registry (attribute-discovered + DI descriptors):
  (default) → CloudAdapter   (via GetDefault → IChannelAdapter)
  "a2a"     → A2AAdapter      ([ChannelAdapter("a2a")])
  "msteams" → TeamsAdapter    ([ChannelAdapter("msteams")])
```

Existing code that resolves `IAgentHttpAdapter` from DI continues to get the
default AP adapter (CloudAdapter). The registry is additive infrastructure.

---

## Summary of Changes

| Component | Change | Breaking? |
|-----------|--------|-----------|
| `IChannelAdapterRegistry` | New interface in **Builder layer** (`Microsoft.Agents.Builder.App`), returns `IChannelAdapter` | No (additive) |
| `ChannelAdapterRegistry` | New implementation (Hosting.AspNetCore); merges attribute + DI descriptors, lazy `ActivatorUtilities` resolution + caching | No (additive) |
| `ChannelAdapterRegistration` | New public descriptor record `(string ChannelId, Type AdapterType)` | No (additive) |
| `ChannelAdapterAttribute` | New class-level `[ChannelAdapter("channelId")]` | No (additive) |
| `ChannelAdapterInitAssemblyAttribute` | New assembly-level attribute + `GetRegistrations()` discovery | No (additive) |
| `ChannelAdapterInitSourceGenerator` | New Roslyn generator in `Microsoft.Agents.Core.Analyzers` (ships in `Microsoft.Agents.Core` package) | No (additive) |
| `AgentApplicationOptions.ChannelAdapterRegistry` | New property exposing the registry (populated from DI) | No (additive) |
| `MapAgentApplicationEndpoints` | Tier 2 resolution (peek channelId, cast to `IAgentHttpAdapter`, fall back to default) | No (existing behavior preserved) |
| `AddCloudAdapter<T>()` | Registers `IChannelAdapterRegistry` singleton | No (additive) |
| `A2AAdapter` | Annotated `[ChannelAdapter("a2a")]` | No (additive) |
| `IChannelAdapter` / `IAgentHttpAdapter` / `IAgent` | Unchanged | — |
| `AgentInterfaceAttribute` | Already exists — unchanged | — |

---

## Implementation Order

1. **`ChannelAdapterRegistration` record and `IChannelAdapterRegistry` interface**
   (interface in Builder layer, returning `IChannelAdapter`; record + impl in
   Hosting.AspNetCore).
2. **`ChannelAdapterAttribute`** (class-level) and **`ChannelAdapterInitAssemblyAttribute`**
   (assembly-level + `GetRegistrations()`).
3. **`ChannelAdapterInitSourceGenerator`** in `Microsoft.Agents.Core.Analyzers`
   (mirrors `ActivityTypeInitSourceGenerator`).
4. **`ChannelAdapterRegistry`** — merges attribute-discovered registrations with DI
   descriptors; lazy `ActivatorUtilities` resolution + caching.
5. **Register `IChannelAdapterRegistry` singleton during `AddCloudAdapter<T>()`.**
6. **Expose `AgentApplicationOptions.ChannelAdapterRegistry`** (populated from DI ctor).
7. **Update `MapAgentApplicationEndpoints`** for Tier 2: resolve the registry, peek
   channelId when `HasChannelSpecificAdapters`, cast to `IAgentHttpAdapter`, fall back to
   default.
8. **Annotate `A2AAdapter` with `[ChannelAdapter("a2a")]`** and add the analyzer
   ProjectReference to the A2A project.

`AgentInterfaceAttribute` already exists. Existing agents with no channel-specific
adapters continue working unchanged (Tier 2 peek is never triggered).

---

## Relationship to Other Documents

- **[Non-Breaking Multi-Adapter Extension Design](2026-06-10-non-breaking-multi-adapter-addendum.md)** —
  covers the extension pattern, custom ITurnContext, custom handler delegates, route
  builders, ITurnState extension scopes, and developer experience. Depends on this
  registry for adapter dispatch.
- **[Multi-Protocol ITurnContext Design](2026-05-12-multi-protocol-turncontext-design.md)** —
  the full breaking design that this approach incrementally converges toward.


</details>

src/libraries/Hosting/AspNetCore/AgentEndpointExtensions.cs
```diff
@@ -123,8 +126,15 @@ public static IEndpointConventionBuilder MapAgentApplicationEndpoints(
...
                     agentGroup.MapMethods(agentInterface.Path, ["POST"],
-                        async (HttpRequest request, HttpResponse response, IAgentHttpAdapter adapter, IServiceProvider services, CancellationToken cancellationToken) =>
+                        async (HttpRequest request, HttpResponse response, IAgentHttpAdapter defaultAdapter, IChannelAdapterRegistry registry, IServiceProvider services, CancellationToken cancellationToken) =>
                         {
+                            // Tier 2 resolution: when channel-specific adapters are registered, peek the
+                            // channelId from the inbound Activity and resolve a channel-specific adapter.
+                            // Otherwise (the common case) use the default AP adapter (CloudAdapter) directly.
+                            var adapter = registry != null && registry.HasChannelSpecificAdapters
+                                ? await ResolveAdapterAsync(registry, defaultAdapter, request, cancellationToken).ConfigureAwait(false)
+                                : defaultAdapter;
```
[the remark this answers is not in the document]
tracyboehrer: Fixed in c1882934. The minimal-API delegate no longer requires `IChannelAdapterRegistry`; it resolves the optional registry from the injected `IServiceProvider` and falls back directly to the registered `IAgentHttpAdapter` when absent. Added a regression test proving the no-registry path returns the default adapter without reading the request body.

src/libraries/Builder/Microsoft.Agents.Builder/App/Proactive/Proactive.cs
```diff
@@ -42,6 +43,42 @@ internal Proactive(AgentApplication app)
...
         }
 
+        /// <summary>
+        /// Resolves the <see cref="IChannelAdapter"/> registered for the specified <paramref name="channelId"/> using
+        /// the <see cref="AgentApplicationOptions.ChannelAdapterRegistry"/>.
+        /// </summary>
+        /// <param name="channelId">The channelId whose adapter should be resolved. Cannot be null or empty.</param>
+        /// <returns>The adapter registered for the channel.</returns>
+        /// <exception cref="InvalidOperationException">Thrown if no registry is available, or no adapter is registered for the channel.</exception>
+        private IChannelAdapter ResolveAdapter(string channelId)
```
ceciliaavila, line 53: Would it make sense for this method to fall back to the default adapter when no channel-specific adapter is configured, rather than failing? This could simplify the use of SendActivityAsync, since callers wouldn't need to be aware of the adapter configuration for a given channel. I'm not sure whether the current behavior is by design.

tracyboehrer: What is the default adapter?  The registry would fallback to a default.  In the normal case this would be CloudAdapter since that is what our service extensions do (AddAgent ultimately adds the defaults bits).

But say someone bypasses that and wires up DI by hand.  It would be tedious, but technically possible.  This could be the case if someone isn't useing Microsoft.Agents.Hosting.AspNetCore.  And say they forgot to register the ChannelRegistry. 

I could see if there is a path to automatically have a default registry, but the question would be what adapter?

tracyboehrer: The default is actually not always CloudAdapter.  It will be the one found by the registered IChannelAdater.

What we could do is: In AgentApplicationOptions construciton a IChannelAdapterRegistry wasn't found, we could create a default ChannelRegistry that contains the registered IChannelAdapter.  In this way, there would always be a IChannelRegistry which contain the DI'd adapter.  But that ONLY happens if that constructor is used.

Definitely limits to what we can do if someone goes to these lengths.

ceciliaavila: Oh, got it. My concern was mainly around the CloudAdapter fallback scenario, but it sounds like that's already handled by the registry. Thanks for clarifying!

ceciliaavila (approved):

Issue #442, opened 2025-09-12
Migration question: strategy for Scoped services


LuisM000:
Hi!
We’re migrating a large Bot Framework solution (using TeamsActivityHandler) to the Microsoft 365 Agents SDK. Our code relies heavily on Scoped services across many layers.

As part of the migration, we’re switching to the new CloudAdapter, which enqueues tasks. I understand that `HostedActivityService` does not resolve `IAgent` within a scope, which could cause issues since it won’t share the same request scope.
However, we still need a safe way to use Scoped services per “turn.”

**Questions:**

1. What’s the recommended migration approach for solutions that rely heavily on Scoped services?
2. Will there be a “synchronous” adapter option (similar to Bot Framework) that processes inline, without using the background queue, to make migration smoother?

Thanks!

EXT97937:
same problem here, any solution? @tracyboehrer??

gioce90:
We have the same problem. Migrating to the new singleton CloudAdapter is really messy. Someone have any solution or suggestion?

gioce90:
I made these changes and it seems works fine:


In my Program.cs: 
```
builder.AddMyBot();
```

Then in WebApplicationBuilderExtensions.cs, you can add your agent/bot as the default recommendation, but there are some overrides of some internal methods of Agents SDK library:

```
public static WebApplicationBuilder AddMyBot(this WebApplicationBuilder builder)
{

    builder.Services.AddScoped<StartDialog>();

    builder.AddAgentApplicationOptionsCustom(); ; // a "scoped" version of AddAgentApplicationOptions() method
    builder.AddAgentCustom<MyBot<StartDialog>, CustomAdapter>(); // a "scoped" version of AddAgent() method

    // Magic trick: remove HostedActivityService in DI
    var toRemove = builder.Services.Where(
        d => d.ServiceType == typeof(IHostedService)
        && d.ImplementationType?.FullName == "Microsoft.Agents.Hosting.AspNetCore.BackgroundQueue.HostedActivityService").ToList();
    foreach (var d in toRemove)
        builder.Services.Remove(d);
    builder.Services.AddHostedService<ScopedHostedActivityService>();

    builder.Services.AddStateServices();

    return builder;
}

    #region Override AddAgent registration

    /// <summary>
    /// Registers AgentApplicationOptions for AgentApplication-based Agents.
    /// </summary>
    /// <remarks>
[...]

PR #960, opened 2026-08-04
Add opt-in per-turn DI scope to HostedActivityService (#442)

LaithZdevnas:
## Problem

`HostedActivityService.ProcessAsync` resolves the Agent from the **root** `IServiceProvider`:

```csharp
var agent = _serviceProvider.GetService(activityWithClaims.AgentType ?? typeof(IAgent));
agent ??= _serviceProvider.GetService(typeof(IAgent));
```

The comment above it correctly identifies that the request scope is already disposed by this point. But resolving from root doesn't avoid scoping — it silently changes the lifetime of everything scoped underneath the Agent. From `CallSiteRuntimeResolver` in dotnet/runtime:

```csharp
protected override object? VisitScopeCache(ServiceCallSite callSite, RuntimeResolverContext context)
{
    // Check if we are in the situation where scoped service was promoted to singleton
    // and we need to lock the root
    return context.Scope.IsRootScope ?
        VisitRootCache(callSite, context) :
        VisitCache(callSite, context, context.Scope, RuntimeResolverLock.Scope);
}
```

`VisitRootCache` caches the instance on the call site and registers it with `Root.CaptureDisposable(...)`, so it lives until the `ServiceProvider` is disposed at host shutdown. For an app that wrote `builder.Services.AddDbContext<T>()` — scoped by default — that means **one `DbContext` shared by every concurrent turn for the lifetime of the process**, with a change tracker that never releases anything.

Two things make this hard to discover:

1. `ValidateScopes` defaults to `HostingEnvironment.IsDevelopment()`, so the invalid resolution throws locally and succeeds in production.
2. When it does throw, `ProcessAsync` catches it and logs `"Error occurred executing WorkItem."`. The Agent simply stops replying, with nothing pointing at DI.

This is #442, open since September 2025 and reported by several teams migrating from Bot Framework.

## Prior art in this repo

The SDK already implements per-turn scoping — in `Hosting/DirectLine.NamedPipes/NamedPipeActivityHandler.cs`:

```csharp
using var scope = _services.CreateScope();
var adapter = scope.ServiceProvider.GetRequiredService<IChannelAdapter>();
var agent = scope.ServiceProvider.GetRequiredService<IAgent>();
...
await adapter.ProcessActivityAsync(identity, activity, agent.OnTurnAsync, cancellationToken);
```

That's currently the only `CreateScope` in `src/libraries`. This PR brings the ASP.NET host in line with it.

## Change

Adds `AdapterOptions.UseScopePerTurn`, **default `false`** so existing behavior is unchanged. When enabled, each queued Activity resolves its Agent from a scope that is disposed once the turn completes:

```csharp
using var turnScope = _useScopePerTurn ? _serviceProvider.CreateScope() : null;
var turnServices = turnScope?.ServiceProvider ?? _serviceProvider;

var agent = turnServices.GetService(activityWithClaims.AgentType ?? typeof(IAgent));
agent ??= turnServices.GetService(typeof(IAgent));
```

No constructor change — `_serviceProvider` is already injected. `activityWithClaims.ChannelAdapter` is captured at enqueue time and is a singleton, so nothing else moves.

## Risk

Low. `grep -rn "AddScoped" src/` returns no registrations — everything the SDK registers is a singleton, and singletons resolve identically from a child scope. Enabling the option can only affect registrations made by the application itself, which today are silently broken anyway.

## Tests

Two tests added to `HostedActivityServiceTests`, using a real `ServiceCollection` rather than the mocked `IServiceProvider` the existing tests use, so scoping is actually exercised:

- `ExecuteAsync_WithScopePerTurn_ShouldResolveScopedDependencyPerTurn` — two turns get distinct instances, and each is disposed after its turn.
- `ExecuteAsync_WithoutScopePerTurn_ShouldShareScopedDependencyAcrossTurns` — asserts `Assert.Same` across turns and that the instance is never disposed, pinning today's behavior so the difference is explicit.

`Microsoft.Agents.Hosting.AspNetCore.Tests`: 85/85 passing. Full `AgentSdk.proj` build succeeds with no new warnings.

## Notes

Happy to adjust naming, or to flip the default to `true` if the team would rather this be the standard behavior — that would also stop transient `IDisposable` Agents accumulating in the root scope, since `CaptureDisposable` captures regardless of lifetime and the root scope only drains at shutdown.

A related follow-up I'd be glad to send separately: `ActivityTaskQueue`, `BackgroundTaskQueue`, `HostedActivityService` and `HostedTaskService` are all `internal`, so the workaround people are using today in #442 has to string-match `"Microsoft.Agents.Hosting.AspNetCore.BackgroundQueue.HostedActivityService"` to remove it from DI. Making these public/overridable would let people subclass instead.

### DIFF
--- a/src/libraries/Hosting/AspNetCore/AdapterOptions.cs
+++ b/src/libraries/Hosting/AspNetCore/AdapterOptions.cs
@@ -34,7 +34,10 @@ namespace Microsoft.Agents.Hosting.AspNetCore
         /// from the root <see cref="System.IServiceProvider"/>. Any scoped registration in the Agent's dependency graph is then
         /// promoted to the root scope, giving a single instance shared by every turn for the lifetime of the process.
         /// Set this to <see langword="true"/> for Agents that depend on scoped services, such as an Entity Framework Core
-        /// <c>DbContext</c>. Services registered by the SDK are singletons and resolve identically either way.
+        /// <c>DbContext</c>. The SDK registers no scoped services, so enabling this affects only registrations made by the
+        /// application. Note that disposable transient dependencies resolved for a turn - <see cref="Microsoft.Agents.Builder.IAgent"/>
+        /// itself is registered transient - are then disposed with the turn scope, rather than being retained by the root
+        /// scope until the host shuts down.
         /// </remarks>
         public bool UseScopePerTurn { get; set; } = false;
     }
--- a/src/libraries/Hosting/AspNetCore/AdapterOptions.cs
+++ b/src/libraries/Hosting/AspNetCore/AdapterOptions.cs
@@ -25,5 +25,17 @@ namespace Microsoft.Agents.Hosting.AspNetCore
         /// Gets or sets a value indicating whether Activity.ServiceUrl should be validated using the 'serviceurl' claim in the incoming token. This is typically used to ensure that the request is coming from a trusted source.
         /// </summary>
         public bool ValidateServiceUrl { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets a value indicating whether each queued Activity is processed in its own dependency injection scope.
+        /// </summary>
+        /// <remarks>
+        /// When <see langword="false"/> (the default), queued Activities resolve the <see cref="Microsoft.Agents.Builder.IAgent"/>
+        /// from the root <see cref="System.IServiceProvider"/>. Any scoped registration in the Agent's dependency graph is then
+        /// promoted to the root scope, giving a single instance shared by every turn for the lifetime of the process.
+        /// Set this to <see langword="true"/> for Agents that depend on scoped services, such as an Entity Framework Core
+        /// <c>DbContext</c>. Services registered by the SDK are singletons and resolve identically either way.
+        /// </remarks>
+        public bool UseScopePerTurn { get; set; } = false;
     }
 }
--- a/src/libraries/Hosting/AspNetCore/BackgroundQueue/HostedActivityService.cs
+++ b/src/libraries/Hosting/AspNetCore/BackgroundQueue/HostedActivityService.cs
@@ -54,8 +54,13 @@ namespace Microsoft.Agents.Hosting.AspNetCore.BackgroundQueue
             ArgumentNullException.ThrowIfNull(activityTaskQueue);
             ArgumentNullException.ThrowIfNull(provider);
 
-            _shutdownTimeoutSeconds = options != null ? options.ShutdownTimeoutSeconds : 60;
-            _useScopePerTurn = options != null && options.UseScopePerTurn;
+            // AdapterOptions is not registered in DI, so `options` is null unless the application registers it.
+            // Fall back to the same "CloudAdapterOptions" configuration section CloudAdapter binds, so the options
+            // are honored from appsettings as this constructor's `config` parameter documents.
+            var adapterOptions = options ?? config.GetSection("CloudAdapterOptions")?.Get<AdapterOptions>() ?? new AdapterOptions();
+
+            _shutdownTimeoutSeconds = adapterOptions.ShutdownTimeoutSeconds;
+            _useScopePerTurn = adapterOptions.UseScopePerTurn;
             _activityQueue = activityTaskQueue;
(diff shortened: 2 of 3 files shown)


LaithZdevnas:

LaithZdevnas:
> @LaithZdevnas please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

LaithZdevnas:
Thanks — all three review comments were correct and are addressed in 0676e5e9.

**1. `AdapterOptions` not bound from configuration.** This was the important one: the option was unreachable as written. `AdapterOptions` isn't registered in DI anywhere in the SDK (`grep -rn "AdapterOptions" src/libraries/ | grep -i "addsingleton\|addtransient\|addscoped\|Configure<"` returns nothing), so the optional `options` parameter is always `null` in practice and `UseScopePerTurn` could never be `true`.

It also surfaced a pre-existing issue on the same line. The constructor takes `IConfiguration` and its own parameter docs say it is *"used to retrieve ShutdownTimeoutSeconds from appsettings"* — but the parameter was never read, so `ShutdownTimeoutSeconds` was pinned at 60 regardless of what `CloudAdapterOptions` contained. Now bound from the same section `CloudAdapter` already uses:

```csharp
var adapterOptions = options ?? config.GetSection("CloudAdapterOptions")?.Get<AdapterOptions>() ?? new AdapterOptions();

_shutdownTimeoutSeconds = adapterOptions.ShutdownTimeoutSeconds;
_useScopePerTurn = adapterOptions.UseScopePerTurn;
```

Worth flagging explicitly since it's a behavior change beyond the new flag: anyone who has `CloudAdapterOptions:ShutdownTimeoutSeconds` set in appsettings will now have it applied to the hosted service's drain window, where previously it was silently ignored. That's the documented intent, but it is a change. Happy to split it into its own PR if you'd rather keep this one strictly additive.

(`HostedTaskService` has the same `AdapterOptions options = null` parameter and no `IConfiguration` at all, so its `ShutdownTimeoutSeconds` is still always 60. Left alone as out of scope — glad to follow up separately.)

**2 & 3. "Everything the SDK registers is a singleton" was inaccurate.** Correct — `AddAgent` registers `IAgent` via `services.AddTransient<IAgent, TAgent>()`. The accurate claim is that the SDK registers no *scoped* services (`grep -rn "AddScoped" src/` returns no registrations). Both the XML docs and the inline comment now say that, and both note the disposal consequence you raised: disposable transients resolved for a turn are disposed with the turn scope instead of being retained by the root scope until host shutdown. That's a deliberate improvement — `CaptureDisposable` captures regardless of lifetime, so today a disposable transient Agent accumulates in the root scope once per turn — but it is a semantic change and should be stated rather than implied.

**Tests:** added `ExecuteAsync_WithScopePerTurnFromConfiguration_ShouldResolveScopedDependencyPerTurn`, which enables the flag through `CloudAdapterOptions:UseScopePerTurn` rather than the constructor parameter, so the path applications will actually use is covered.

86/86 passing in `Microsoft.Agents.Hosting.AspNetCore.Tests`, full `AgentSdk.proj` build clean.

tracyboehrer:
Closing this in favor of #974, which incorporates and expands the scoped-service fix here.

A huge thank you to @LaithZdevnas for identifying the root-provider lifetime issue, documenting the production impact so thoroughly, and contributing the original implementation. Your work directly informed the replacement and its test coverage.

PR #994, opened 2026-09-02
Return unsupported for extended agent cards

tracyboehrer:
## Summary

- explicitly advertise extended agent cards as unsupported in the generated agent card
- return `UnsupportedOperationError` (`-32004`) from the custom Agents SDK JSON-RPC path
- add integration coverage for the public agent card and extended-card JSON-RPC request

## Context

The Agents SDK maps its own A2A endpoints and uses local HTTP/JSON-RPC processors rather than the a2a-dotnet ASP.NET Core endpoints. Those processors delegate to an `A2AServer`, whose current preview2 implementation returns `ExtendedAgentCardNotConfiguredError` (`-32007`) even though this integration does not advertise the capability. The adapter-specific server override keeps the custom endpoint consistent with its generated agent card.

The corresponding upstream server fix is a2aproject/a2a-dotnet#475.

This PR is stacked on #698 because the A2A extension has not yet merged to `main`. Its diff against `users/tracyboehrer/a2a-dotnet-int` contains only this compliance fix.

## TCK failure

`tests/compatibility/core_operations/test_error_handling.py::TestCapabilityExtendedCard::test_extended_card_not_supported_jsonrpc`

```text
CORE-CAP-003 [Extended agent card operations return error when not supported] failed on jsonrpc: Expected error code -32004 (UnsupportedOperationError), got -32007
```

## Tests

- `dotnet test src/tests/Microsoft.Agents.Extensions.A2A.Tests/Microsoft.Agents.Extensions.A2A.Tests.csproj --no-restore --nologo`

### DIFF
--- a/src/libraries/Extensions/Microsoft.Agents.Extensions.A2A/A2AAdapter.cs
+++ b/src/libraries/Extensions/Microsoft.Agents.Extensions.A2A/A2AAdapter.cs
@@ -104,6 +104,7 @@ public class A2AAdapter : ChannelAdapter, IA2AHttpAdapter
             Skills = [],
             Capabilities = new AgentCapabilities()
             {
+                ExtendedAgentCard = false,
                 Streaming = true,
             },
             SupportedInterfaces = [],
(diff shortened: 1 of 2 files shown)


tracyboehrer:
This fix has been applied directly to the integration branch in commit c3a2edca and is now part of PR #698.

PR #1013, opened 2026-09-14
Fix Slack feedback channel deserialization

goldenwitch-at-msft:
# Fix Slack feedback channel deserialization

## Summary

The Slack sample's `-stream` feedback buttons produce an object-valued `channel`. Azure Bot Service already delivers the `message/submitAction` Activity, but accessing `SlackActivity.ChannelData` throws `JsonException` at `$.Payload.channel` because `ActionPayload.channel` is a string.

- Accept `channel.id` and legacy string channels without changing the public property type.
- Fall back to `container.channel_id`; allow missing channel context and reject malformed values.
- Replay a sanitized serialized Activity through `ProtocolJsonSerializer` and `SlackFeedbackLoopRoute`, asserting one handler call with the expected reaction, channel, and thread.

Object-valued channels normalize to the string ID in the typed model; serializing `ActionPayload` emits that string, not the original channel object. Existing thread resolution and top-level extension data are unchanged.

This SDK fix does not require an Azure Bot Service deployment. The fixture uses deterministic placeholders based on the documented [block_actions](https://docs.slack.dev/reference/interaction-payloads/block_actions-payload/) and [feedback_buttons](https://docs.slack.dev/reference/block-kit/block-elements/feedback-buttons-element/) shapes, not a real-user transcript.

## Validation

```powershell
dotnet build src/tests/Microsoft.Agents.Extensions.Slack.Tests/Microsoft.Agents.Extensions.Slack.Tests.csproj
dotnet test src/tests/Microsoft.Agents.Extensions.Slack.Tests/Microsoft.Agents.Extensions.Slack.Tests.csproj -f net8.0
dotnet test src/tests/Microsoft.Agents.Extensions.Slack.Tests/Microsoft.Agents.Extensions.Slack.Tests.csproj -f net10.0
```

Local validation on September 14, 2026: all targets build with zero warnings and errors; 180 tests pass on each of .NET 8 and .NET 10.

.NET Framework 4.8 compiles, but its tests remain unexecuted because the local runner rejects delay-signed assemblies. Live Slack end-to-end validation is not claimed.


goldenwitch-at-msft (commented):
Contract proof and minimal-implementation review of `1d902e14`; findings are inline.

src/libraries/Extensions/Microsoft.Agents.Extensions.Slack/Api/ActionPayload.cs
```diff
@@ -18,4 +21,36 @@ public class ActionPayload : SlackModel
...
+            {
+                return null;
+            }
+
+            if (channelId.ValueKind != JsonValueKind.String)
+            {
+                throw new JsonException("Slack channel id must be a string.");
+            }
+
+            return channelId.GetString();
```
goldenwitch-at-msft, line 50: **Medium: full Activity serialization loses native channel fields.** This conversion is also used by `ProtocolJsonSerializer.ToJson(slackActivity)` through the `SlackActivity.ChannelData` shadow. With this PR's fixture, the emitted `channelData.Payload.channel` changes from `{ "id": "D0LOCALDEV01", "name": "directmessage" }` to `"D0LOCALDEV01"`; `Payload.Get<string>("channel.name")` is null. A nested unknown channel member is likewise discarded. I reproduced this against the PR head. The raw base `Activity.ChannelData` stays intact, so checking it or top-level `AdditionalProperties` alone misses the loss in the serialized Activity.

Please preserve the channel object for native-payload access and Activity round-trips, deriving only the convenience channel ID, and add an object-channel round-trip assertion to the existing Activity tests. The public string compatibility choice should not silently change the full Activity's wire shape.

src/tests/Microsoft.Agents.Extensions.Slack.Tests/AgentApplicationAttributesTests.cs
```diff
@@ -272,6 +273,23 @@ public async Task SlackFeedbackLoopRoute_FiresOnFeedback()
...
+    {
+        var app = new FeedbackLoopRouteApp(new AgentApplicationOptions((IStorage)null));
+        var activity = ProtocolJsonSerializer.ToObject<Activity>(SlackActivityTests.FeedbackActivityJson);
+        var turnContext = CreateTurnContext(activity);
+        Assert.Equal("1700000000.000000", ProtocolJsonSerializer.ToObject<FeedbackData>(activity.Value).ReplyToId);
+
+        await app.OnTurnAsync(turnContext.Object, CancellationToken.None);
+
+        Assert.Equal("OnFeedback", Assert.Single(app.calls));
+        Assert.Equal("positive_feedback", Assert.Single(app.feedback).ActionValue.Reaction);
```
goldenwitch-at-msft, line 287: **Medium: the regression proves route selection, not the sample's reply.** This test passes with the committed fixture even though `channelData.ApiToken` is absent; the nested `Payload.token` is a different field. The sample's `OnSlackFeedbackLoopAsync` passes `channelData.ApiToken` to `SlackApi.CallAsync`. Replaying that call with this fixture and a recording HTTP handler emits `/api/chat.postMessage` without an Authorization header. This is a gap in the proof that the sample feedback flow completes, not a newly introduced outbound SDK bug.

The minimal stronger proof would extend this existing fixture and route test with a sanitized `ApiToken` and a recording `SlackApi`, following the existing `SlackApiTests.CreateFactory` pattern. Assert exactly one outbound request with the expected bearer token, channel, thread, and reaction text. That exercises the sample's response boundary without requiring live Slack or additional service/production changes.

src/libraries/Extensions/Microsoft.Agents.Extensions.Slack/Api/ActionPayload.cs
goldenwitch-at-msft: On independent re-review, I am withdrawing this as a required change. The normalization is real, including when serializing a SlackActivity, but lossless re-emission of the newly accepted channel object is not part of the acceptance contract for this fix. The PR explicitly keeps the public string property and normalizes object input to its ID. The existing tests prove object/string channel resolution, precedence, container fallback, thread resolution, and the existing top-level extension-data behavior; the raw base Activity still retains the native payload. Requiring preservation of arbitrary nested channel fields would add a representation and output contract rather than pin an unmet acceptance criterion. No additional production change or regression test is needed for this scoped fix. Resolving with no change.

src/tests/Microsoft.Agents.Extensions.Slack.Tests/AgentApplicationAttributesTests.cs
goldenwitch-at-msft: On independent re-review, I am withdrawing the additional test request. The scoped acceptance boundary is successful deserialization and exactly one SlackFeedbackLoopRoute handler invocation with the expected reaction, channel, and thread; SlackFeedbackLoopRoute_FiresForSerializedFeedbackActivity already proves that path through OnTurnAsync. This fixture is intentionally a deserialization/routing fixture, not a credential-bearing outbound replay. Azure Bot Service supplies ApiToken separately, and neither that injection nor SlackApi authentication changes in this PR. A synthetic-token chat.postMessage capture would be optional integration coverage, not necessary proof of this fix. The PR already disclaims live Slack end-to-end validation. Resolving with no change so the PR remains minimal.

goldenwitch-at-msft:
Superseded by https://github.com/microsoft/Agents-for-net/pull/1014, which uses an upstream branch as requested by the maintainers. The head commit is unchanged: 1d902e142face2994394776015d5058ac48cbfc7. Existing reviewer requests were carried over; these resolved self-review discussions and their rationale remain here for reference. Please continue review on #1014.

